PR #7, opened 2020-04-26
Docker refactor

zees-dev:
- modified directory structure to better suit docker-compose based solution
- created custom config files to be mounted to docker-compose server containers
- moved sql file to more relevant location; run upon container startup (once only)
- ignore Data.nx in current folder (mounted at container startup)


docker/Dockerfile
```diff
@@ -1,11 +1,18 @@
...
+RUN chmod +x Valhalla
 
-ADD . $SRC_DIR
+RUN echo "#!/bin/sh" > docker-entrypoint.sh && \
+  echo "set -ex" >> docker-entrypoint.sh && \
+  echo 'exec "$@"' >> docker-entrypoint.sh && \
+  chmod +x docker-entrypoint.sh
 
-RUN cd $SRC_DIR; go build -o server
+EXPOSE 8484 8585 8584 8685
```
Hucaru, line 15: No need to expose ports here as the compose overrides.

zees-dev:
Will update PR soon.

docker/Dockerfile
zees-dev: Its best practice to expose ports in Dockerfile so that one can understand what ports an app exposes without going through any app code.
I will expose relevant ports however.

zees-dev:
The master branch is broken - issue raised. Unfortunately i did rebase or branch off the master. I'll wait until that is resolved, then rebase and update this PR.

docker/Dockerfile
Hucaru: I agree with the statement, but in this case the app exposes ports based on config which isn't baked into the image. Therefore, I feel it is best for a user to manually expose the ports they have determined via compose or standard docker command line.

zees-dev:
Pull request updated.

Also tidied go mod "go mod tidy"

docker/Dockerfile
zees-dev: Makes sense, will update accordingly :)

PR #9, opened 2020-08-22
Added function to clean up database and reset dangling users.

ErwinsExpertise:


server/login.go
```diff
@@ -35,6 +35,20 @@ func (server *LoginServer) Initialise(dbuser, dbpassword, dbaddress, dbport, dbd
...
 	log.Println("Connected to database")
+
+	server.CleanupDB()
+
+	log.Println("Cleaned up the database")
+}
+
+// CleanupDB sets all accounts isLogedIn to 0
+func (server *LoginServer) CleanupDB() {
+	res, err := server.db.Exec("UPDATE `accounts` SET `isLogedIn`=0 WHERE `isLogedIn` = 1;")
```
Hucaru, line 46: I think currently this would allow users to login multiple times. I suspect would need to check that characters for this account have a `channelID` of -1 and a `migrationID` of -1 and only then reset the account?

server/login.go
```diff
@@ -35,6 +35,20 @@ func (server *LoginServer) Initialise(dbuser, dbpassword, dbaddress, dbport, dbd
...
+	if err != nil {
+		log.Println(err)
```
Hucaru, line 48: Should probably terminate the application as this is a pretty bad error

server/login.go
ErwinsExpertise: I have noticed sometimes the account will become stuck and maintain a migrationID of 0 if kicked out at the character selection screen; however, the channelID should be -1 for all characters on account regardless of migrationID for any accounts that are currently not logged in. I can add in the check for channelID on all characters but the migrationID is not necessarily something that can be checked for the purpose of this function in its current state.

server/login.go
ErwinsExpertise: Changed to log.Fatal()

server/login.go
Hucaru: Good point. In that case just adding check for the channelID being -1 should be all that is needed.

PR #11, opened 2020-09-01
Added metrics port flag and refactored metrics initialization

ErwinsExpertise:
Current setup for metrics will throw errors on a single-host non-dockerized environment. Added flag to specify metrics port. Default for metrics port is 9000 if not specified so the Docker-Compose configuration will work without issue.

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -3,14 +3,13 @@ package metrics
 import (
 	"log"
 	"net/http"
-	"strconv"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 // Port metrics are served on
-const Port = 9000
+var Port string
 
 var (
 	// Gauges available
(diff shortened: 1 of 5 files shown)


server/metrics/metrics.go
```diff
@@ -20,12 +19,15 @@ var (
 	Counters = make(map[string]*prometheus.CounterVec)
 )
 
-func init() {
+// StartMetrics initializes and handles metrics Prometheus endpoint
+func StartMetrics() {
```
Hucaru, line 23: Why change this to an explicit call instead of package init? My understanding (not tested) is that you can initialise Prometheus http handler and metrics independently.

ErwinsExpertise: This was needed because init is ran when the file is accessed or prior to any other function. Due to the way the server is setup a metric endpoint was being set up for each server(login, world, and all channels). During testing there were some apparent race conditions appearing due to the use of init rather than an explicit call. This was primarily that the metric Port variable was not able to get set prior to the initialization of the endpoint. By making this an explicit call rather than using init we are able to control the initialization of the endpoint and prevent race conditions.

PR #12, opened 2020-09-02
Began implementing mob damage

ErwinsExpertise:
Implementing mob damage. Still need to work on mob skills, however, this implementation will allow for mob melee damage at a minimum. This code was pulled from commit 4320cdb and adjusted to work with current structure.

### DIFF
--- a/server/channel_player.go
+++ b/server/channel_player.go
@@ -5,6 +5,8 @@ import (
 	"log"
 	"strconv"
 
+	"github.com/Hucaru/Valhalla/server/player"
+
 	"github.com/Hucaru/Valhalla/constant"
 	"github.com/Hucaru/Valhalla/constant/opcode"
 	"github.com/Hucaru/Valhalla/mnet"
@@ -13,7 +15,6 @@ import (
 	"github.com/Hucaru/Valhalla/server/message"
 	"github.com/Hucaru/Valhalla/server/metrics"
 	"github.com/Hucaru/Valhalla/server/movement"
-	"github.com/Hucaru/Valhalla/server/player"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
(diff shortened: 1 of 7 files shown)


server/channel_player.go
```diff
@@ -648,3 +650,93 @@ func (server ChannelServer) playerMoveInventoryItem(conn mnet.Client, reader mpa
...
+
+		// Magic guard dmg absorption
+
+		// Fighter / Page power guard
+
+		// Meso guard
+
+		plr.GiveHP(int16(playerDamage))
+
+		inst.Send(player.PlayerReceivedDmg(plr.ID(), mobAttack, damage, reducedDamage, spawnID, mobID, healSkillID, stance, reflectAction, reflected, reflectX, reflectY))
```
Hucaru, line 736: Could this be wrapped within the player package? Maybe as `GiveHP(dmg int16, show bool)` (might have situations where negative hp might not need to be shown?) as the player has a ptr to the instance they are in. This also keeps the `PlayerReceiveDmg function` in package scope only and gives it the same nomenclature as the other packet functions.

ErwinsExpertise: Definitely. I will get that setup and amend the PR. I agree that having it scoped to the player package would be more effective and helpful down the line.

server/player/player.go
```diff
@@ -1058,3 +1058,19 @@ func (d Data) Save(db *sql.DB) error {
...
+	}
+	newHP := d.hp - damage
+
+	if newHP <= -1 {
+		d.hp = 0
+	} else {
+		d.hp = newHP
+	}
+
+	d.Send(packetPlayerStatChange(show, constant.HpID, int32(d.hp)))
```
Hucaru, line 1075: The packet function has the bool at the front listed as unknown, Have you tested to see what it does?

ErwinsExpertise: I did test that, but I did not see any difference for melee or bump damage when it was provided true or false. I had forgotten to remove it from the function. Removed and amended

PR #21, opened 2020-09-05
Building out class structs & constants

ErwinsExpertise:


server/channel_mob.go
```diff
@@ -138,31 +138,12 @@ func (server ChannelServer) mobDamagePlayer(conn mnet.Client, reader mpacket.Rea
 
 }
 
-func (server ChannelServer) mobEffect(conn mnet.Client, reader mpacket.Reader) {
+func (server ChannelServer) mobDistanceFromBoss(conn mnet.Client, reader mpacket.Reader) {
 	/*
 		id := reader.ReadInt32()
 		attack := reader.ReadInt32()
```
Hucaru, line 144: Should name this distance as we roughly know what it is.

constant/skill/class.go
```diff
@@ -17,3 +17,277 @@ type gm struct {
 
 type superGm struct {
```
Hucaru, line 18: Can delete this and the other un-exported classes

constant/constants.go
```diff
@@ -117,3 +117,199 @@ var ExpTable = [...]int32{15, 34, 57, 92, 135, 372, 560, 840, 1242, 1144, // Beg
 	894634784, 943660770, 995373379, 1049919840, 1107455447, 1168144006, 1232158297, 1299680571,
 	1370903066, 1446028554, 1525246918, 1608855764, 1697021059, // 0 is the amount of exp needed for level 200 to level up i.e. never shall
 }
+
+const (
```
Hucaru, line 121: If we made these constants of type Skill we could write a `String() string` method that does a map lookup to return the human readable skill name for debugging and logging.

Hucaru:
As a general thing are you working on mobs and skills?

server/channel_mob.go
ErwinsExpertise: I believe this packet only applies to bosses. I found it in WvsGlobal and they do not seem to do anything with it either:
https://github.com/diamondo25/WvsGlobal/blob/master/WvsBeta.Game/Packets/MobPacket.cs#L220

constant/skill/class.go
ErwinsExpertise: Perfect, I was unsure if they were being use elsewhere already.

ErwinsExpertise:
Primarily mob skills at the moment. I just noticed that the classes were not built out so I am adding them to have later on. 

Also, I believe it would be best to change the variables to global variables so I will also adjust the naming convention of those variables.

constant/skill/class.go
```diff
@@ -17,3 +17,277 @@ type gm struct {
...
+	Resurrection    Skill
+	SuperDragonRoar Skill
+	Teleport        Skill
+
+	ItemExplosion Skill
+	ShadowPartner Skill
+	JumpDown      Skill
+}
+
+var (
```
ErwinsExpertise, line 267: I believe it would be better to have the variables as global vars and simply init them in the package. 

Ex.

```
var Swordsman swordsman
func init() {
    Swordsman = swordsman{
      JobID: WarriorJobID,
      ImprovedMaxHpIncrease:  ImprovedMaxHpIncrease,
      etc..
    }
}
```

Hucaru: Yea that makes sense. 

I'm not exactly sure how the class skills will end up being used - I wrote the file near the start of the project a while ago. Do you have any plans on how to use these structs?

ErwinsExpertise: I believe they will really just be used as conditionals in an easy to read format. They might be of use in other ways, but I haven't thought of others yet. 

Example:

```
if ( Assassin.Haste || Bandit.Haste) {
 Increase Speed
}
```

ErwinsExpertise:
Once this passes build checks I am good to merge if there is nothing else.

Issue #23, opened 2020-09-08
NPC scripting


Hucaru:
@ErwinsExpertise I have made progress on a script system for NPCs - the scripts are written in JavaScript. Is the following script clear?:

```
var state = 0
var styles = [31050, 31040, 31000, 31060, 31090, 31020, 31130, 31120, 31140, 31330, 31010]
var goods = [ [1332020],[1332020, 1],[1332009, 0] ]

function run(npc, player) {
	if (npc.next()) {
		state++
	} else if (npc.back()) {
		state--
	}

	if (state == 2) {
		if (npc.yes()) {
			state = 3
		} else if (npc.no()) {
			state = 4
		}
	} else if (state == 4) {
		if (npc.selection() == 1) {
			state = 0
		} else if (npc.selection() == 2) {
			state = 5
		} else if (npc.selection() == 3) {
			state = 6
		} else if (npc.selection() == 4) {
			state = 7
		} else if (npc.selection() == 5) {
			state = 8
		}
	}

	switch(state) {
	case 0:
		npc.sendBackNext("first", false, true)
		break
	case 1:
		npc.sendBackNext("second", true, false)
		break
	case 2:
		npc.sendYesNo("finished")
		break
	case 3:
		npc.sendOK("selection:" + npc.selection() + ", input number:" + npc.inputNumber() + ", input text: " + npc.inputString())
		npc.terminate()
		break
	case 4:
		npc.sendSelection("Select from one of the following:\r\n#L1#Back to start #l\r\n#L2#Styles#l\r\n#L3#Input number#l#L4#Input text#l\r\n#L5#Shop#l")
		break
	case 5:
		npc.sendStyles("Select from the following", styles)
		state = 3
		break
	case 6:
		npc.sendInputNumber("Input a number:", 100, 0, 100)
		state = 3
		break
	case 7:
		npc.sendInputText("Input text:", "default", 0, 100)
		state = 3
		break
	case 8:
		npc.sendShop(goods)
		break
	default:
		npc.sendOK("state " + state)
		npc.terminate()
	}
}
```

ErwinsExpertise:
Will the source be moving from gomacro to JS? Yes, it does make sense and it is structured well. Will state be required for scripting or is that just for this test script?

Hucaru:
Yea, I will be moving from gomacro to JS. 

> Will state be required for scripting or is that just for this test script?

State management will be up to the writer to implement. For example if you have a shop script you could implement it in just:

```
[...]

PR #24, opened 2020-09-08
Feature/npc scripting

Hucaru:
Scripts load when server starts (static folder), updates to files are automatically detected and reloaded

#23


server/channel_chat.go
```diff
@@ -859,6 +859,19 @@ func (server *ChannelServer) gmCommand(conn mnet.Client, msg string) {
 		inst, err := field.GetInstance(plr.InstanceID())
 		pool := inst.DropPool()
 		pool.RemoveDrop(false, id)
+	case "npco":
```
Hucaru, line 862: I think there is a system to override npc script behaviour e.g. cody not sure how it works in the slightest at the moment.

zees-dev (commented):
With my very limited knowledge i have made a comment/suggestion for the JS scripts in this PR.

I have very limited knowledge of the codebase but am very excited with all the recent activity!

I would like to get involved but have no clue how how you reverse engineer the client and/or work with IDA. On a side note, any resources/guidance here would be appreciated.

Great job guys!

scripts/npc/1012000.js
```diff
@@ -0,0 +1,53 @@
+// Henesys Regular Cab
+
+var towns = [104000000, 102000000, 101000000, 103000000]
+var prices_text = ["800", "1,000", "1,000", "1,200"]
+var prices_num = [800, 1000, 1000, 1200]
+
+var state = 0
+var selection = -1
```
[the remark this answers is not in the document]
Hucaru: Unfortunately the VM doesn't support `const` and `let`.

Each script has it's own global namespace so cross-script pollution should not occur.

We can't define state within the `run` function as run gets executed for each stage of the script, meaning state would be reset upon each execution (I might have misunderstood what you are describing here, if so an example would be appreciated as I am not that familiar with js).

The scripts are run in the following manner:

1. Compile to byte code (this will generate and populate the global namespace)
2. call `run`
4. Read script action and set any flags e.g. yes, no, next, back, selection
5. goto 2

Hucaru:
> With my very limited knowledge i have made a comment/suggestion for the JS scripts in this PR.
> 
> I have very limited knowledge of the codebase but am very excited with all the recent activity!
> 
> I would like to get involved but have no clue how how you reverse engineer the client and/or work with IDA. On a side note, any resources/guidance here would be appreciated.
> 
> Great job guys!

1. First get a copy of IDA: https://www.hex-rays.com/products/ida/support/download_freeware/ (the free version should be ok)
2. Download the idb from: https://hostr.co/3NYvnemS3cAJ

You can open the idb with IDA and on the left you should see a list of named and unnamed (sub_****) subroutines. Double click a named subroutine and press f5 will give c++ decompiled code. It can look nasty sometimes as IDA will do a literal translation between the assembly and c/c++.

If you see Decode1, Decode2 etc function calls they are packet read calls letting you know that the subroutine you are in is reading a packet.

YouTube and google are good resources for learning more on IDA

server/channel_chat.go
ErwinsExpertise: I believe this is controlled in the string node of the wz file. Possibly leveraging the data in that node can be used to quickly create GMS-like scripts as well?

scripts/npc/1012000.js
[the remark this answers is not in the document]
zees-dev: o wow, i see; ok the current approach makes sense if that is the case!

server/channel_chat.go
Hucaru: I have had a look at that and I think the nodes under `/String/Npc.img/` are responsible for the bubble dialogue above an npc and default text that you get when double clicking them.

ErwinsExpertise: Right. I think I may have misunderstood the behavior you are describing.

ErwinsExpertise:
> With my very limited knowledge i have made a comment/suggestion for the JS scripts in this PR.
> 
> I have very limited knowledge of the codebase but am very excited with all the recent activity!
> 
> I would like to get involved but have no clue how how you reverse engineer the client and/or work with IDA. On a side note, any resources/guidance here would be appreciated.
> 
> Great job guys!

Always great to have more help! I would love to see this project take off. There are a lot of aspects that do not require much IDB digging but still need to be worked out. Thing such as guild, parties, chat(whisper, etc) have most of the opcodes already defined in the source and just need to be implemented. Player skills as well, however, there might be some packets that the server needs to send to apply stats from the skill.

PR #26, opened 2020-09-10
Adding mob skills

ErwinsExpertise:
This process seems to be rather drawn out so I will be implementing this in sections. 

Process to use skill:
1. Load skills from NX
2. Identify if mob can use skill/currently using
3. If eligible to use skill assign a random skill as next use

Once an eligible skill has been identified it must be applied. Depending on the type of skill this can be various different methods. 
Most common:
Apply Stats (Seal, Stun, etc)
Summon Mobs

Both methods will likely be independent PR's as well. 

At the moment this code will assign the next skill to use. It does need to be cleaned up a little bit. Currently, when applying the skill to the controller it is getting stuck in a loop. Is there a default skillID to set for regular attacks to bring it back?

Continuing to work on this.

### DIFF
--- a/server/channel_mob.go
+++ b/server/channel_mob.go
@@ -1,8 +1,13 @@
 package server
 
 import (
+	"time"
+
+	skills "github.com/Hucaru/Valhalla/constant/skill"
+
 	"github.com/Hucaru/Valhalla/mnet"
 	"github.com/Hucaru/Valhalla/mpacket"
+	"github.com/Hucaru/Valhalla/nx"
 	"github.com/Hucaru/Valhalla/server/field/lifepool/mob"
 	"github.com/Hucaru/Valhalla/server/movement"
 	"github.com/Hucaru/Valhalla/server/player"
@@ -17,20 +22,26 @@ func (server ChannelServer) mobControl(conn mnet.Client, reader mpacket.Reader) 
 
 	skillPossible := (bits & 0x0F) != 0
 
-	plr, err := server.players.getFromConn(conn)
+	var actualAction int8
 
+	if action < 0 {
+		actualAction = -1
+	} else {
+		actualAction = action >> 1
+	}
+
+	plr, err := server.players.getFromConn(conn)
 	if err != nil {
 		return
 	}
 
-	field, ok := server.fields[plr.MapID()]
-
-	if !ok {
+	inst, err := server.getPlayerInstance(conn, reader)
+	if err != nil {
 		return
 	}
 
-	inst, err := field.GetInstance(plr.InstanceID())
-
+	var mob mob.Data
+	mob, err = inst.LifePool().GetMobFromID(mobSpawnID)
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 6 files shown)


server/channel_player.go
```diff
@@ -688,3 +688,26 @@ func (server ChannelServer) playerBumpDamage(conn mnet.Client, reader mpacket.Re
 	plr.DamagePlayer(int16(damage))
 
 }
+
+func (server *ChannelServer) getPlayerInstance(conn mnet.Client, reader mpacket.Reader) (*field.Instance, error) {
```
Hucaru, line 692: This should probably be `(server ChannelServer)` as this function shouldn't modify ChannelServer

server/channel_mob.go
```diff
@@ -39,21 +50,52 @@ func (server ChannelServer) mobControl(conn mnet.Client, reader mpacket.Reader)
...
+
+	inst.LifePool().MobAcknowledge(mobSpawnID, plr, moveID, skillPossible, byte(action), skillData, moveData, finalData, moveBytes)
 
-	// if actualAction >= 21 && actualAction <= 25 {
-	// 	mob.PerformSkill(skillDelay, skillLevel, skillID)
-	// } else if actualAction > 12 && actualAction < 20 {
-	// 	mob.PerformAttack(byte(actualAction - 12))
-	// }
+	// This will update the next skill but currently in a loop
+	//mob.AcknowledgeController(moveID, finalData, skillPossible, skillID, skillLevel)
```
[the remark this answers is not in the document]
Hucaru: What do you mean by "it is in a loop"? Does the mob do the same skills straight away after finishing the previous one?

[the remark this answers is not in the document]
ErwinsExpertise: Sorry let me explain more. 

Once a mob begins attacking it will find a random skill for that mob to use. The skill will then be applied to the mob and it will use that skill. For the moment I have the skill cooldown set at 10 seconds so during those 10 seconds the function on line 91 returns 0,0 until the cooldown is reached. Once cooldown has been reached then it will get a new skill.

During the 10 seconds while the function is returning 0,0 the mob uses the same skill that it was given at the beginning. It does not revert itself back to "normal" attacks during this time, and will just constantly use that same skill over and over until a new one is given. Ideally, what I want to do is use the skill, and then after the skill is used set the mob back to normal attacks.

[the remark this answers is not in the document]
Hucaru: That seems odd as I think that is controlled by `skillPossible` and `skillID` variables. When I had it previously implemented the mob was definitely respecting the skills usage cooldown. I suspect the false is more important than the skill level, id and mp from looking at the packet: 

https://github.com/Hucaru/Valhalla/blob/master/server/field/lifepool/mob/packet.go#L23-L28

I wouldn't be surprised if the client reads the bool and if it's false stops reading the packet. Other things to try would be setting the mp to zero and see if that forces it to stop. You could also try inverting the boolean value as it might not technically be a boolean (0 false, 1 true) in the client.

> function on line 91 returns 0,0 until the cooldown is reached

Have you verified this is actually happening with a print statement or otherwise?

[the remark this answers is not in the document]
Hucaru: I think I know where the bug is. skillPossible doesn't observe the timer. So everytime the controller tells the server I want a skill (line 24) it will always hit the else block. Even though you return 0,0 the skillPossible flag is true.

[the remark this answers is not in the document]
ErwinsExpertise: Yes, I have verified it is return the proper variables. I believe I found the issue. I was previously sending the values to 2 different controllers. I updated it to create a new skilldata variable and update with the proper data. It seems to be working correctly now.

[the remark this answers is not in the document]
Hucaru: I edited my previous comment with an issue with the function in general.

server/field/lifepool/pool.go
```diff
@@ -263,22 +263,50 @@ func (pool *Data) NpcAcknowledge(poolID int32, plr controller, data []byte) {
...
+						if mob.MP() < 0 {
+							mob.SetMP(0)
+						}
+					}
+
+				}
+
+				pool.mobs[i].PerformAttack(attackID)
 			}
+			mob.SetSkillPossible(skillPossible)
```
Hucaru, line 307: I'm not sure if this is the correct behaviour. The way I see it is as follows:
1. Controller sends a request to use a skill (skillPossible)
2. Server then selects an off-cooldown skill
3a. If a skill is possible then skillPossible is set to true and a skillID and level is sent
3b. If a skills is not possible then skillPossible is set to false and skillID and level are set to zero

I think in the 3b scenario the bool will be set to true in the packet as skillPossible is being passed through but not changed? I think `CanUseSkill` should take a boolean and return bool, byte, byte where the boolean is either true or false depending on if a skill is selected.

Note: I am not sure if my initial assumption in point 1 is correct.

ErwinsExpertise: I agree with this. This was the method I was initially envisioning as well, and I will refactor the function to act accordingly.

ErwinsExpertise: I don't think is a need to set skillPossible at all actually. I think this variable is only needed within the scope of choosing a skill. Otherwise the client should let us know when a skill is possible again. It seems to be working fine in new implementation.

Hucaru: I have checked out the PR locally and mobs appear to be acting properly. Will you be setting the skill cooldowns to the values defined in nx? Will you add the mob mp regen in this PR?

ErwinsExpertise: Yes cooldowns will be based on the values in nx file. Have not looked much into mp regen, but I can try to implement that as well this PR.

server/field/lifepool/mob/mob.go
```diff
@@ -471,6 +477,25 @@ func (m Data) String() string {
...
+
+	if m.poison {
+		// Handle poison
+		m.hp = m.hp - 10 // Need to adjust to poison amount based on poison level
+	}
+
+	// Update mob status if one is applied
+	if (checkTime - m.lastHeal) > 30 {
+		// Heal the mob
+		regenhp, regenmp := int32(10), int32(10) // Need to figure out how to find these values? Are they set in nx file for mob?
```
ErwinsExpertise, line 495: I've looked around but I can't find what these values should be? I can make an algorithm if nothing else. Maybe heal 0.01% or so every 30 seconds? I am not too sure what these numbers should look like.

Hucaru: No problem, might have to look at other sources for that. I didn't know mobs could heal hp passively!

When you are ready for this branch to be merged in leave a comment.

ErwinsExpertise: Alright sounds good. I will continue to dig on it, and if I do not find anything I will just put together a simple algorithm for now. 

I believe the HP heal might be a bit more difficult to implement if I cannot find values. Obviously we don't want the mob healing if it is actively being attacked so I will need to add checks for that as well. MP should be rather simple because it should not matter whether or not it is being attacked.

Hucaru: I think for hp it should be based on time since it was last attacked? Maybe have something like: After 60 seconds of not being attacked hp is restored to full? I can't remember if the original game had hp regen?

ErwinsExpertise: I think full hp after 60 seconds might be too aggressive. It would be not great especially for bosses or high hp mobs. I've set it to restore 1% after 60 seconds every 30 seconds. I believe this should be a fair heal amount for the time being.

ErwinsExpertise:
If you are comfortable with the merge it should be good to merge.

server/field/lifepool/mob/mob.go
```diff
@@ -63,14 +66,19 @@ type Data struct {
 	link                   int32
 	flySpeed               int32
 	noRegen                int32
-	skills                 map[byte]byte
+	Skills                 map[byte]byte
```
Hucaru, line 69: Should this be private?

ErwinsExpertise: This variable needed to be global for server/field/lifepool/pool.go #288 to work. I can make a global function instead that returns the Skills map.

PR #30, opened 2020-10-02
Implementing basic useable items

ErwinsExpertise:
This will remove the item after consumption, and apply any HP or MP if loaded into the data.


server/channel_npc.go
```diff
@@ -228,11 +228,9 @@ func (server *ChannelServer) npcShop(conn mnet.Client, reader mpacket.Reader) {
 			return
 		}
 
-		_, err = plr.TakeItem(itemID, amount)
+		invID := byte(2) // Need to adjust this to find the inventory ID for selling
```
Hucaru, line 231: First digit is the inventory id: 1 = equip, 2 = use, 3 = set-up, 4 = etc, all others = cash

Divide the item id by 1000000 to get the value

ErwinsExpertise: Oh perfect. I was not aware of that structure. I will update it to find the inventory ID that way.

server/player/player.go
```diff
@@ -442,6 +442,39 @@ func (d *Data) SetMapPosID(pos byte) {
...
+}
+
+// UseItem of Data
+func (d *Data) UseItem(slot int16, id int32) {
+	if d.hp < 1 {
+		d.noChange()
+		return
+	}
+
+	item, err := d.getItem(2, slot)
```
Hucaru, line 456: You will need to do the same here as well I suspect.

ErwinsExpertise: That shouldn't be necessary if the id structure does not stray. This function is solely for usable items which should always be in the use inventory.

server/channel_player.go
```diff
@@ -661,6 +661,19 @@ func (server ChannelServer) playerMoveInventoryItem(conn mnet.Client, reader mpa
...
+func (server ChannelServer) playerUseInventoryItem(conn mnet.Client, reader mpacket.Reader) {
+	plr, err := server.players.getFromConn(conn)
+	if err != nil {
+		return
+	}
+
+	slot := reader.ReadInt16()
+	itemid := reader.ReadInt32()
+
+	plr.UseItem(slot, itemid)
```
Hucaru, line 673: I think the item usage should be something like:
```
item, err := plr.TakeItem()

if err != nil {
}

item.Apply(plr)
```

This way item logic would be within the item package instead of bloating the player package.

ErwinsExpertise: I like the idea of making more reusable code; however, wouldn't that create an import cycle? Specifically creating a function in the item package that accepts type *player.Data . I believe that would be needed to alter the players stats when an item is applied.

Hucaru: The `Apply` function should take an interface instead of the `*player.Data` type. This has the added benefit of making it easier to unit test items if that ever gets done, as the unit test would create a dummy type that matches the interface.

ErwinsExpertise: I'm not sure I understand what you are wanting out of the item.Apply function. If an interface is passed rather than a struct we will not be able to interact with the player's data. 

As an example for the following function
`func (v *Data) ApplyUseItem(plr interface{}) {
}`

We would not be able to do something such as this to apply an HP potion because an interface is ambiguous therefore we cannot make the call to plr.hp . 
`	if v.hp > 0 {
		if (plr.hp - v.hp) < 0 {
			plr.hp = 0
		} else {
			plr.hp = plr.hp - v.hp
		}
	}`

Can you explain a little more what you are envisioning specifically for the item.Apply() function?

Hucaru: What I mean by interface is something like:
```
type player interface {
  NoChange()
  GiveHP(int16)
  GiveMP(int16)
}

func (v *Data) ApplyUseItem(plr player) { }
```

instance.go in the field package does a similar thing - at the top of the file you will see a player interface defined

ErwinsExpertise: Ah okay perfect, that will work. I will refactor it and update PR.

server/channel_player.go
```diff
@@ -670,7 +670,7 @@ func (server ChannelServer) playerUseInventoryItem(conn mnet.Client, reader mpac
 	slot := reader.ReadInt16()
 	itemid := reader.ReadInt32()
 
-	plr.UseItem(slot, itemid)
+	plr.ApplyUseItem(slot, itemid, server.db)
```
Hucaru, line 673: I don't think plr needs this function. You could replace this line with the contents of the function as the level of abstraction it provides isn't needed.

ErwinsExpertise: I agree, it isn't really needed. I removed and amended the commit.

server/player/player.go
```diff
@@ -670,13 +641,25 @@ func (d *Data) GiveItem(newItem item.Data, db *sql.DB) error { // TODO: Refactor
...
+
+	maxRemove := math.Min(float64(item.Amount()), float64(amount))
+
+	item.UpdateAmount(item.Amount() - int16(maxRemove))
+
+	if item.Amount() == 0 {
+		// Delete item
+		d.removeItem(item, db)
+	} else {
+		//Update item with new amount
```
Hucaru, line 658: `packetInventoryAddItem` will probably do what yo want here.

server/player/player.go
```diff
@@ -670,13 +641,25 @@ func (d *Data) GiveItem(newItem item.Data, db *sql.DB) error { // TODO: Refactor
 }
 
 // TakeItem from Data
-func (d *Data) TakeItem(id int32, slot int16, amount int16, invID byte) {
-
+func (d *Data) TakeItem(id int32, slot int16, amount int16, invID byte, db *sql.DB) item.Data {
 	item, err := d.getItem(invID, slot)
 	if err != nil {
-		return
+		log.Println(err)
```
Hucaru, line 647: I think the `item.ID() == id` check should be here instead of channel_player.go. If either of these checks fail the error should be returned and handled by the calling function.

server/player/player.go
ErwinsExpertise: That packet works great. I need to implement the DB side of things and then I'll push that commit into this PR.

ErwinsExpertise: I am not sure why but the update is not persisting past a single use. For example if you use 1 potion it updates fine, but if you use 3 in a row it will still act as if it is one use. So rather than 10 -> 7 for 3 potions in a row it will be 10 -> 9. The db is updating after usage as well, but because the item in memory is not updating past the single use the DB is not getting updated correctly either. 

I have reviewed the code and I am not seeing where this issue could be occurring at. Do you have any insight on this?

Hucaru: `getItem` is returning a copy of the item, you need a ptr.

ErwinsExpertise: I was able to get it working by using the item Save and updateItem functions.

server/player/player.go
```diff
@@ -667,8 +678,9 @@ func (d *Data) TakeItem(id int32, slot int16, amount int16, invID byte, db *sql.
 
 }
 
-func (d *Data) updateItemStack(item item.Data, db *sql.DB) {
-	item.UpdateStack(db)
+func (d Data) updateItemStack(item item.Data, db *sql.DB) {
+	item.Save(db, d.id)
+	d.updateItem(item)
```
Hucaru, line 683: Passing the database ptr around everywhere is getting kind of messy, I think it would suit being a global of it's own package (done as another PR)?

ErwinsExpertise: I agree completely. It would be best to have the database as its own package. I think that would also help with some syncing issues as well. I also agree that it is something that would need to be a separate PR also it can be tracked easier that way.

Hucaru:
Is this PR ready to merge in? I am not in a position to test this at the moment.

ErwinsExpertise:
Yes, it can be merged.

PR #51, opened 2022-06-22
Added player drops

ErwinsExpertise:
Adding player drops. Item drops working as expected, however, meso drop essentially freezes player. I believe the code is getting hung up somewhere and it is being suppressed.

### DIFF
--- a/channel/handlers.go
+++ b/channel/handlers.go
@@ -62,6 +62,8 @@ func (server *Server) HandleClientPacket(conn mnet.Client, reader mpacket.Reader
 		server.npcShop(conn, reader)
 	case opcode.RecvChannelInvMoveItem:
 		server.playerMoveInventoryItem(conn, reader)
+	case opcode.RecvChannelPlayerDropMesos:
+		server.playerDropMesos(conn, reader)
 	case opcode.RecvChannelInvUseItem:
 		server.playerUseInventoryItem(conn, reader)
 	case opcode.RecvChannelAddStatPoint:
@@ -721,6 +723,21 @@ func (server Server) playerMoveInventoryItem(conn mnet.Client, reader mpacket.Re
 	if err != nil {
 		log.Println(err)
 	}
+}
+
+func (server Server) playerDropMesos(conn mnet.Client, reader mpacket.Reader) {
+	amount := reader.ReadInt32()
+	plr, err := server.players.getFromConn(conn)
+
+	if err != nil {
+		return
+	}
+
+	err = plr.dropMesos(amount)
+	if err != nil {
+		log.Println(err)
+	}
+
 }
 
 func (server Server) playerUseInventoryItem(conn mnet.Client, reader mpacket.Reader) {
--- a/channel/player.go
+++ b/channel/player.go
@@ -1,6 +1,7 @@
 package channel
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"math"
@@ -396,6 +397,10 @@ func (d *player) giveMesos(amount int32) {
 	d.setMesos(d.mesos + amount)
 }
 
+func (d *player) takeMesos(amount int32) {
+	d.setMesos(d.mesos - amount)
+}
+
 // UpdateMovement - update Data from position data
 func (d *player) UpdateMovement(frag movementFrag) {
 	d.pos.x = frag.x
@@ -773,6 +778,17 @@ func (d *player) removeItem(item item) {
 	d.send(packetInventoryRemoveItem(item))
 }
 
+func (d *player) dropMesos(amount int32) error {
+	if d.mesos < amount {
+		return errors.New("not enough mesos")
+	}
+
+	d.takeMesos(amount)
+	d.inst.dropPool.createDrop(dropSpawnNormal, dropFreeForAll, amount, d.pos, true, 0, 0)
+
+	return nil
+}
+
 func (d *player) moveItem(start, end, amount int16, invID byte) error {
 	if end == 0 { //drop item
 		fmt.Println("Drop item amount:", amount)
@@ -783,7 +799,8 @@ func (d *player) moveItem(start, end, amount int16, invID byte) error {
 		}
 
 		d.removeItem(item)
-		// inst.AddDrop()
+
+		d.inst.dropPool.createDrop(dropSpawnNormal, dropFreeForAll, 0, d.pos, true, d.id, 0, item)
 	} else if end < 0 { // Move to equip slot
 		item1, err := d.getItem(invID, start)
 
--- a/channel/pools.go
+++ b/channel/pools.go
@@ -841,32 +841,34 @@ func (pool *dropPool) createDrop(spawnType byte, dropType byte, mesos int32, dro
 		timeoutTime = currentTime.Add(itemLootableByAllTimeout).Unix()
 	}
 
-	for i, item := range items {
-		tmp := dropFrom
-		tmp.x = dropFrom.x - offset + int16(i*itemDistance)
-		finalPos := pool.instance.calculateFinalDropPos(tmp)
+	if len(items) > 0 {
+		for i, item := range items {
+			tmp := dropFrom
+			tmp.x = dropFrom.x - offset + int16(i*itemDistance)
+			finalPos := pool.instance.calculateFinalDropPos(tmp)
 
-		if poolID, err := pool.nextID(); err == nil {
-			drop := fieldDrop{
-				ID:      poolID,
-				ownerID: ownerID,
-				partyID: partyID,
-				mesos:   0,
-				item:    item,
+			if poolID, err := pool.nextID(); err == nil {
+				drop := fieldDrop{
+					ID:      poolID,
+					ownerID: ownerID,
+					partyID: partyID,
+					mesos:   0,
+					item:    item,
 
-				expireTime:  expireTime,
-				timeoutTime: timeoutTime,
-				neverExpire: false,
+					expireTime:  expireTime,
+					timeoutTime: timeoutTime,
+					neverExpire: false,
 
-				originPos: dropFrom,
-				finalPos:  finalPos,
+					originPos: dropFrom,
+					finalPos:  finalPos,
 
-				dropType: dropType,
+					dropType: dropType,
+				}
+
+				pool.drops[drop.ID] = drop
+
+				pool.instance.send(packetShowDrop(spawnType, drop))
 			}
-
-			pool.drops[drop.ID] = drop
-
-			pool.instance.send(packetShowDrop(spawnType, drop))
 		}
 	}
 
--- a/unkown-packet.log
+++ b/unkown-packet.log
@@ -1,0 +1,11 @@
+Finishing quest with Sara:
+UNKNOWN CLIENT PACKET: [Packet] (12) : 47 02 E8 03 34 08 00 00 FF FF FF FF
+
+Accept quest with Olaf in Lith
+UNKNOWN CLIENT PACKET: [Packet] (8) : 47 01 21 08 75 4A 0F 00
+
+Pickup korean fan
+[Packet] (9) : 73 22 03 B7 FF 00 00 00 00
+
+Drop 10 mesos
+UNKNOWN CLIENT PACKET: [Packet] (5) : 3C 0A 00 00 00
--- a/common/opcode/recv.go
+++ b/common/opcode/recv.go
@@ -34,6 +34,7 @@ const (
 	RecvChannelPassiveRegen        byte = 0x37
 	RecvChannelAddSkillPoint       byte = 0x38
 	RecvChannelSpecialSkill        byte = 0x39
+	RecvChannelPlayerDropMesos     byte = 0x3C
 	RecvChannelCharacterInfo       byte = 0x3F
 	RecvChannelLieDetectorResult   byte = 0x45
 	RecvChannelCharacterReport     byte = 0x49
--- a/config_channel_3.toml
+++ b/config_channel_3.toml
@@ -1,0 +1,18 @@
+[database]
+address = "127.0.0.1"
+port = "3306"
+user = "root"
+password = "password"
+database = "maplestory"
+
+[channel]
+worldAddress = "127.0.0.1"
+worldPort = "8584"
+listenAddress = "0.0.0.0"
+listenPort = "8683"
+ClientConnectionAddress = "127.0.0.1"
+packetQueueSize = 512
+MaxPop = 250
+# The following should be set to zero when not testing on a local network environment
+latency = 0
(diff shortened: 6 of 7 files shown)


ErwinsExpertise:
Relevant vana code: https://github.com/retep998/Vana/blob/19116c07dd0082b6c570bfd6d2ea604947773c10/src/channel_server/drop_handler.cpp#L198

ErwinsExpertise:
Quantity drop is not 100% working but pickup/drop for mesos is 100% and single items/full stacks is working

Hucaru:
I tested this locally to test and found two issues:
1. The message for the amount of mesos received is not correct
2. When dropping an item and then picking it up followed by a relog, the item is gone from inventory, if using /drop command and picking up the items it spawns then they get saved, apart from the mesos, picking that up doesn't change the value in inventory.

channel/handlers.go
```diff
@@ -723,6 +727,21 @@ func (server Server) playerMoveInventoryItem(conn mnet.Client, reader mpacket.Re
...
 
+func (server Server) playerDropMesos(conn mnet.Client, reader mpacket.Reader) {
+	amount := reader.ReadInt32()
+	plr, err := server.players.getFromConn(conn)
+
+	if err != nil {
+		return
+	}
+
+	err = plr.dropMesos(amount)
```
Hucaru, line 738: I don't think the player struct should know about the concept of dropping mesos. This handler should deduct the mesos from the player and then create a drop based on the player id and mesos value i.e. the 'flow' should probably be like:
```
|-handler
      |-> player remove mesos
      | -> pool add drop
```

instead of:

```
|- handler
       |- > player remove mesos
               | -> pool add drop
```

channel/player.go
```diff
@@ -773,8 +779,56 @@ func (d *player) removeItem(item item) {
...
+	if d.pos.x-pos.x > 800 || d.pos.y-pos.y > 600 {
+		// Hax
+		log.Printf("player: %s tried to pickup an item from far away", d.name)
+		d.inst.send(packetDropNotAvailable())
+		d.inst.send(packetInventoryDontTake())
+		return
+	}
+
+	if drop.mesos > 0 {
+		//mozes
```
Hucaru, line 810: Comment not needed. I assume it's supposed to be mesos?

channel/pools.go
```diff
@@ -813,8 +815,40 @@ func (pool *dropPool) eraseDrops() {
 	pool.drops = make(map[int32]fieldDrop)
 }
 
-func (pool *dropPool) playerAttemptPickup(dropID int32, position pos) (bool, item) {
-	return false, item{}
+func (pool *dropPool) getPickupDrop(dropID int32) (error, fieldDrop) {
+	drop := pool.findDropFromID(dropID)
+
+	if reflect.DeepEqual(drop, fieldDrop{}) {
```
Hucaru, line 821: reflect.DeepEqual is expensive. Update findDropFromID to return the drop and and ok value.

channel/pools.go
```diff
@@ -813,8 +815,40 @@ func (pool *dropPool) eraseDrops() {
...
+
+	var amount int16
+
+	if drop.mesos > 0 {
+		amount = int16(drop.mesos)
+	} else {
+		amount = drop.item.amount
+	}
+
+	if err := pool.instance.send(packetPickupNotice(drop.item.id, amount, drop.mesos > 0, drop.item.invID == 1.0)); err != nil {
```
Hucaru, line 841: I don't think this should be sent to the whole pool

channel/pools.go
```diff
@@ -813,8 +815,40 @@ func (pool *dropPool) eraseDrops() {
...
+	drop := pool.findDropFromID(dropID)
+
+	if reflect.DeepEqual(drop, fieldDrop{}) {
+		return errors.New("unavailable"), fieldDrop{}
+	}
+
+	return nil, drop
+}
+
+func (pool *dropPool) playerAttemptPickup(drop fieldDrop, playerID int32) error {
```
Hucaru, line 828: pass the player directly so you can send them specific packets

channel/player.go
```diff
@@ -160,6 +161,7 @@ func (d player) send(packet mpacket.Packet) {
 	}
 
 	d.conn.Send(packet)
+	d.save()
```
Hucaru, line 164: Not a good idea to save to database every time a packet is sent as database saving is a slow operation. We should only save at very specific points to minimise the number of writes and the size of the update (currently the save function saves everything and needs to be broken down into categories such as inventory saves, character avatar saves, friends list saves, guild saves etc).

unkown-packet.log
```diff
@@ -0,0 +1,17 @@
+Finishing quest with Sara:
```
Hucaru, line 1: Prefer placing these debug/analysis comments in the relevant .go files to help keep things together in one place

ErwinsExpertise:
@Hucaru thank you for the hook, that will definitely help. 

I appreciate you taking time to review the PR.

As far as the meal amount goes that's an issue I haven't been able to figure out. The packet is structured basically exactly like the Vana pickup notice yet the meso amount is not showing. I know you're a bit busy but if you have time could you review that specific packet? I'm really not sure what I'm doing wrong with that one.

Everything else I'll get taken care of.

channel/player.go
ErwinsExpertise: That makes sense. I can remove this and work on adding specific saves.

channel/pools.go
```diff
@@ -1051,10 +1075,50 @@ func packetShowDrop(spawnType byte, drop fieldDrop) mpacket.Packet {
...
 	p := mpacket.CreateWithOpcode(opcode.SendChannelDropExitMap)
-	p.WriteBool(instant) // 0 - fade away, 1 - instant, 2,3,5 - player id? , 4 - int16
+	p.WriteInt8(dropType) // 0 - fade away, 1 - instant, 2 - loot by user, 3 - loot by mob
 	p.WriteInt32(dropID)
+	p.WriteInt32(lootedBy)
+
+	return p
+}
+
+func packetPickupNotice(itemID int32, amount int16, isMesos bool, isEquip bool) mpacket.Packet {
```
Hucaru, line 1087: From what I can tell the packet structure is as follows:
```
func packetPickupNotice(itemID int32, amount int16, isMesos bool, isEquip bool, cafeBonus int16) mpacket.Packet {
	p := mpacket.CreateWithOpcode(opcode.SendChannelInfoMessage)
	p.WriteInt8(0) //??

	p.WriteBool(isMesos)

	if isMesos {
		p.WriteInt32(int32(amount))
		p.WriteInt16(cafeBonus)
	} else {
		p.WriteInt32(itemID)

		if !isEquip {
			p.WriteInt16(amount)
		}

		p.WriteInt32(0)
	}

	return p
}
```

ErwinsExpertise: That worked perfectly, ah when I tried before I didn't cast the amount as int32, oops. Thank you for it. I made the changes.

Hucaru:
I think the dropping of items that have an amount greater than 1 such as potions should have the window pop up asking for the amount before the packet is sent, not really sure. It could be there is a property about the item (not the amount?) that tells the client when to do this.

ErwinsExpertise:
I'm not really sure how to get that to pop-up because it does not do it by itself in the client as it would with dropping mesos. The packet itself does not contain any unique identifiers either it simply has (opcode | pos | dropid). I'm assuming it is related to the item itself. I'll review that and see what I can find.

ErwinsExpertise:
I believe they are handling the window via inventory operations here:
https://github.com/retep998/Vana/blob/19116c07dd0082b6c570bfd6d2ea604947773c10/src/channel_server/inventory_handler.cpp#L113

Not too familiar with C++ so it's a little fuzzy but I'll work on it

Hucaru:
I have had another play locally and it appears when dropping a potion and picking it up it can sometimes crash the client and the inventory for that item isn't handled properly i.e. I had a stack of 100 and 200 dropped 100, picked up 100 crashed loged in again and had a stack of 100 only in inventory.

ErwinsExpertise:
I think that has to do with the saving inconsistencies. The server does not save as often as it should so I think it would be best to tackle that as a DB saving refactor. 

Also I believe the dropping issue is relate to the item packets itself as you stated. I read an old ragezone post where it happened because the expiration time wasn't properly set. Will need to review more

ErwinsExpertise:
I resolved the issue with the drop UI not showing by adding an expiration time. Not really sure why that matters, but it does apparently.

channel/item.go
```diff
@@ -391,7 +391,8 @@ func (v item) bytes(shortSlot bool) []byte {
 		p.WriteUint64(uint64(v.id)) // I think this is somekind of cashshop transaction ID for the item
 	}
 
-	p.WriteInt64(v.expireTime)
+	//p.WriteInt64(v.expireTime)
```
Hucaru, line 394: Keep it as the variable and make the default in the item creation the value below.

ErwinsExpertise: changed it to a const and updated so that all items have it by default. Still need to figure out the saving issue.

ErwinsExpertise:
Pretty sure the stacking issue is related to the updateItem function. It looks like the slice is getting copied and we are modifying the copy slice. 
https://github.com/Hucaru/Valhalla/blob/354ba204ae26b73bf43df07b4859f3aac33ac575/channel/player.go#L685

channel/item.go
Hucaru: I might not be a saving issue. The way items are handled when moved/dropped might need to be refactored and simplified as I think that might be what is causing the state of the server and client to differ.

Hucaru:
> Pretty sure the stacking issue is related to the updateItem function. It looks like the slice is getting copied and we are modifying the copy slice.
> 
> https://github.com/Hucaru/Valhalla/blob/354ba204ae26b73bf43df07b4859f3aac33ac575/channel/player.go#L685

That could very well be it. I don't have a specific test for it apart from moving stackable items around and dropping them then logging on and off

ErwinsExpertise:
Drops appear to be working as expected now. The copy slice was getting updated versus the players slice. In addition, when the item was added to the DB item.save function was not using a pointer so it was not updating the actual item, but rather a copy as well.

channel/player.go
```diff
@@ -389,13 +390,35 @@ func (d *player) addEquip(item item) {
 
 func (d *player) setMesos(amount int32) {
 	d.mesos = amount
-	d.send(packetPlayerStatChange(false, constant.MesosID, amount))
+	d.send(packetPlayerStatChange(true, constant.MesosID, amount))
+	d.saveMesos()
```
Hucaru, line 394: I wonder if in the future db write operations should be sent to a dedicated thread to prevent locking up the main thread?

channel/player.go
```diff
@@ -389,13 +390,35 @@ func (d *player) addEquip(item item) {
...
 }
 
+func (d *player) takeMesos(amount int32) {
+	d.setMesos(d.mesos - amount)
+}
+
+func (d *player) saveMesos() error {
+	props := `mesos=?`
+
+	query := "UPDATE characters SET " + props + " WHERE accountID=? and name=?"
```
Hucaru, line 408: Can just inline the mesos props

channel/player.go
```diff
@@ -429,11 +452,35 @@ func (d player) checkPos(pos pos, xRange, yRange int16) bool {
...
 	}
+
+	if err := d.saveMapID(id, oldMapID); err != nil {
+		log.Println(err)
+	}
+
+}
+
+func (d *player) saveMapID(newMapId, oldMapId int32) error {
+	props := `mapID=?,previousMapID=?`
```
Hucaru, line 469: Can also inline these as I don't think it will make the line too long

channel/player.go
```diff
@@ -773,17 +840,61 @@ func (d *player) removeItem(item item) {
...
+	if d.mesos < amount {
+		return errors.New("not enough mesos")
+	}
+
+	d.takeMesos(amount)
+
+	return nil
+}
+
+func (d *player) pickupItem(pos pos, dropID int32) {
```
Hucaru, line 853: The logic in this function is better placed in the handler and doesn't need to be a method.

channel/item.go
```diff
@@ -301,7 +308,7 @@ func (v item) shield() bool {
 	return v.weaponType == 17
 }
 
-func (v item) save(charID int32) (bool, error) {
+func (v *item) save(charID int32) (bool, error) {
```
Hucaru, line 311: Good spot

nx/item.go
```diff
@@ -140,6 +141,55 @@ func extractItems(nodes []gonx.Node, textLookup []string) map[int32]Item {
 		}
 	}
 
+	searches = []string{"/Item/Consume"}
```
Hucaru, line 144: Thanks for adding this

ErwinsExpertise: No problem, it's added some unsupport nx options, but I can get those later.

Hucaru:
It might be worthwhile looking into changing the script system to just be the same as Odin. Have you finished the item related changes you wanted to make? I would like to keep the PRs scoped to a single area/topic/issue if possible.

ErwinsExpertise:
Agreed, it would be a lot easier to mimic Odin versus making 300 different NPC scripts. 

Yes all the changes are done and drops are working exactly as expected now from my testing.

Hucaru:
There is a bug where if an item is spawned e.g. `/item 2000003 100` then dropped and picked up, it is no longer in inventory when reloged (if partial amount dropped and picked up then works as expected). This applies to all item types. Would suggest there is a database saving issue when an item is picked up.

ErwinsExpertise:
> There is a bug where if an item is spawned e.g. `/item 2000003 100` then dropped and picked up, it is no longer in inventory when reloged (if partial amount dropped and picked up then works as expected). This applies to all item types. Would suggest there is a database saving issue when an item is picked up.

Just tested and it looks like this was related to the drop item's DB ID. It kept the same db ID once dropped so changing it to have a db ID of 0 resolved the bug.

channel/pools.go
```diff
@@ -813,8 +814,29 @@ func (pool *dropPool) eraseDrops() {
 	pool.drops = make(map[int32]fieldDrop)
 }
 
-func (pool *dropPool) playerAttemptPickup(dropID int32, position pos) (bool, item) {
-	return false, item{}
+func (pool *dropPool) playerAttemptPickup(drop fieldDrop, player *player) {
+	var amount int16
+
+	player.send(packetRemoveDrop(2, drop.ID, player.id))
```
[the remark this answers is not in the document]
ErwinsExpertise: Fixed

Hucaru:
Happy for me to merge?

ErwinsExpertise:
Yep, I'm working on skills which seems like it's going to be a lot. So this can merge

PR #54, opened 2022-10-04
Feature/pin system

amityahav:
all possible cases where checked and works perfectly i based on these packets which the clients sends:

when registering pin:
09 01 04 00 31 31 31 31

when clicking ok after canceling change pin request: 
09 20

when clicking login on pin window:
08 01 20 01 20 20 20 04 20 31 31 31 31

when clicking change pin on pin window:
08 02 20 20 20 20 20 04 20 31 31 31 31

when clicking cancel on pin window:
08 20

right after login auth success the client sends:
08 01 01 01 20 20 20 20 20

### DIFF
--- a/login/handlers.go
+++ b/login/handlers.go
@@ -21,6 +21,8 @@ func (server *Server) HandleClientPacket(conn mnet.Client, reader mpacket.Reader
 		server.handleLoginRequest(conn, reader)
 	case opcode.RecvLoginCheckLogin:
 		server.handleGoodLogin(conn, reader)
+	case opcode.RecvLoginRegisterPin:
+		server.handlePinRegistration(conn, reader)
 	case opcode.RecvLoginWorldSelect:
 		server.handleWorldSelect(conn, reader)
 	case opcode.RecvLoginChannelSelect:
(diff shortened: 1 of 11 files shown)


login/packets.go
```diff
@@ -33,6 +33,32 @@ func packetLoginResponce(result byte, userID int32, gender byte, isAdmin bool, u
 	return pac
 }
 
+func packetPinOperation(mode byte) mpacket.Packet {
+	pac := mpacket.CreateWithOpcode(opcode.SendLoginPinRegister)
```
Hucaru, line 37: I think a better name for `SendLoginPinRegister` would be `SendLoginPinOperation` as this opcode is used for more than registering.

login/handlers.go
```diff
@@ -76,34 +77,89 @@ func (server *Server) handleLoginRequest(conn mnet.Client, reader mpacket.Reader
...
+		b1 := reader.ReadByte()
+		b2 := reader.ReadByte()
+
+		if b1 == 1 && b2 == 1 { // First attempt, request for pin
+			if len(pinDB) == 0 {
+				conn.Send(packetRegisterPin())
+			} else {
+				conn.Send(packetRequestPin())
+			}
+			return
```
Hucaru, line 131: I'm not a fan of having lots of return statements like this. See if the logic can be done in a way where there aren't as many return statements.

Hucaru (changes requested):

login/handlers.go
amityahav: done

Hucaru:
I'm happy with the changes. If you could update the readme to tick the pin feature I will then merge this in.

Hucaru (approved):

PR #56, opened 2022-10-08
Feature/exp drop rates commands

amityahav:
Hey again, i've implemented the task 'active gm command for change rates for all channels'
this works for all rates with the command:
/rate <exp | drop | mesos> rate
rate is an int16 because its enough.

extra command: /showRates

so the main logic is when gm exec the command it sends a packet to the world channel which stores the new rates in its memory and also propagate the new rates to all channels which then store them also.

when ever a new channel establishes a connection with the world , world will send its rates through the ChannelOk handler, we want this in cases when a new channel registers/channel restarts after GM executed a command and we want the command to take effect.

the world server initially gets its rates from the config file, and if for example it restarts it will override any new rates GM wrote.


world/server.go
```diff
@@ -49,6 +49,10 @@ func (server *Server) ServerDisconnected(conn mnet.Server) {
 	server.login.Send(server.info.GenerateInfoPacket())
 }
 
+func (server *Server) FillRates(exp, drop, mesos int) {
```
Hucaru, line 52: This function is not needed and what it does can just be inline.

channel/commands.go
```diff
@@ -23,6 +24,39 @@ func (server *Server) gmCommand(conn mnet.Client, msg string) {
...
+
+		mode := command[1]
+		mFunc, ok := rates[mode]
+		if !ok {
+			conn.Send(packetMessageRedText("Choose between exp/drop/mesos Rates"))
+			return
+		}
+
+		rate := command[2]
+		r, err := strconv.Atoi(rate)
```
Hucaru, line 47: If this is an integer then it cannot handle rates such `1.5x` or `0.5x`

channel/player.go
```diff
@@ -238,15 +240,16 @@ func (d *player) setEXP(amount int32) {
 }
 
 func (d *player) giveEXP(amount int32, fromMob, fromParty bool) {
+	a := int32(d.rates.exp) * amount
```
Hucaru, line 243: Just re-assign this value to amount no need to make a new variable

channel/handlers.go
```diff
@@ -2452,6 +2455,12 @@ func (server *Server) handleNewChannelBad(conn mnet.Server, reader mpacket.Reade
 func (server *Server) handleNewChannelOK(conn mnet.Server, reader mpacket.Reader) {
 	server.worldName = reader.ReadString(reader.ReadInt16())
 	server.id = reader.ReadByte()
+	server.rates = &Rates{
+		exp:   reader.ReadInt16(),
+		drop:  reader.ReadInt16(),
+		mesos: reader.ReadInt16(),
+	}
+
 	log.Println("Registered as channel", server.id, "on world", server.worldName)
```
Hucaru, line 2462: Should probably output the rates here as well

channel/handlers.go
```diff
@@ -2577,15 +2586,42 @@ func (server *Server) handlePartyEvent(conn mnet.Server, reader mpacket.Reader)
 			party.updateJobLevel(playerID, job, level)
 		}
 	default:
-		log.Println("Unkown party event type:", op)
+		log.Println("Unknown party event type:", op)
+	}
+}
+
+func (server *Server) handleChangeRate(conn mnet.Server, reader mpacket.Reader) {
```
Hucaru, line 2591: This function should also print the new rates to screen

world/handlers.go
```diff
@@ -176,6 +182,26 @@ func (server *Server) handlePartyEvent(conn mnet.Server, reader mpacket.Reader)
 		p.WriteBytes(reader.GetBuffer())
 		server.channelBroadcast(p)
 	default:
-		log.Println("Unkown party event type:", op)
+		log.Println("Unknown party event type:", op)
+	}
+}
+
+func (server *Server) handleChangeRate(conn mnet.Server, reader mpacket.Reader) {
```
Hucaru, line 189: This function should also detect if the rates have gone above `2x` to inform the login server so that it can set the relevant information on world select as I think you can state there is an exp event going on.

https://github.com/Hucaru/Valhalla/blob/master/login/packets.go#L203

You will need to add this to https://github.com/Hucaru/Valhalla/blob/bab971b99f24e989f401d2017a8d5e54d34e301f/internal/serialisable_data.go#L9 
and
 https://github.com/Hucaru/Valhalla/blob/bab971b99f24e989f401d2017a8d5e54d34e301f/internal/serialisable_data.go#L17

Hucaru (changes requested):

Hucaru:
Don't for get to update readme

world/server.go
amityahav: i put this in a method because doing it inline i will need to export the 'info' field when all other fields are not

Hucaru: I'm ok with making the info field exported since all of it's member variables are, but I will leave it up to you.

channel/handlers.go
amityahav: what do you mean 'to screen' ? as a log or to the players? 
cuz a msg to players is sent in another handler

world/handlers.go
amityahav: rates on a private server will probably always be higher then the normal rates. do we want to inform the client like its an exp event?

LuneArk: Whoa, new updates. It's also nice to see a new contributor. :)

May I suggest if rates go higher than `1x`? Exp/Drop events were rare occurrences that happened only a handful of times throughout a year in GMS. I doubt many would play a private server at this low a rate, but it would be nice to see this source as close to the original game.

However, I'm assuming the mentioned function above could just as easily have its value changed to what the server owner considers as the default rates, and anything higher would be considered an Exp/Drop event? If that is the case, then I guess my `1x` suggestion wouldn't matter.

Hucaru: I agree with @LuneArk's last paragraph: the exp rates should be set via a world config file which then distributes it to the channels. Then anything that is set above that value triggers an exp event notification back to the login server so that the event can be displayed on world select

channel/handlers.go
Hucaru: I meant a `println` statement so that it's displayed in the terminal.

world/handlers.go
amityahav: Got you!

amityahav:
updated to pr according to latest comments.

channel/player.go
```diff
@@ -240,16 +240,16 @@ func (d *player) setEXP(amount int32) {
 }
 
 func (d *player) giveEXP(amount int32, fromMob, fromParty bool) {
-	a := int32(d.rates.exp) * amount
+	amount = int32(math.Ceil(float64(d.rates.exp * float32(amount))))
```
Hucaru, line 243: Ceil is not a cheap call to make especially as it gets called every time a players kills a mob or receives exp in general. Just do the standard type cast, this will round down but I think that's ok.

Also this will apply the exp rate to everything the player receives exp for e.g. both quests, PQ stage completion, mob kills. Might want to separate this out?

channel/handlers.go
```diff
@@ -2455,11 +2455,17 @@ func (server *Server) handleNewChannelBad(conn mnet.Server, reader mpacket.Reade
 func (server *Server) handleNewChannelOK(conn mnet.Server, reader mpacket.Reader) {
 	server.worldName = reader.ReadString(reader.ReadInt16())
 	server.id = reader.ReadByte()
-	server.rates.exp = reader.ReadInt16()
-	server.rates.drop = reader.ReadInt16()
-	server.rates.mesos = reader.ReadInt16()
+	// No checking errors because we've validated a couple of times
+	exp, _ := strconv.ParseFloat(reader.ReadString(reader.ReadInt16()), 32)
```
Hucaru, line 2459: Why are we passing the rate values as strings?

internal/packets.go
```diff
@@ -151,22 +151,22 @@ func PacketChannelPartyUpdateInfo(partyID, playerID, job, level int32, name stri
 	return p
 }
 
-func PacketRateOperation(mode byte, rate int16) mpacket.Packet {
+func PacketRateOperation(mode byte, rate string) mpacket.Packet {
 	p := mpacket.CreateInternal(opcode.ChangeRate)
 	p.WriteByte(mode)
-	p.WriteInt16(rate)
+	p.WriteString(rate)
```
Hucaru, line 157: Is it because there is no Write/ReadFloat32/64 method? If so you can implement one. Can't remember the endianess we need for the game but it would be something like:
```
b := *(*uint32)(unsafe.Pointer(&value))
*p = append(byte(b), byte(b >> 8), byte(b >> 16), byte(b >> 24))
```
then do the inverse operation for the read

world/handlers.go
```diff
@@ -199,13 +201,25 @@ func (server *Server) handleChangeRate(conn mnet.Server, reader mpacket.Reader)
...
-		server.info.Rates.Mesos = float32(rs)
+		server.Info.Rates.Mesos = float32(rs)
 	}
 
+	h := sha512.New()
+	h.Write([]byte(fmt.Sprintf("exp %.2f drop %.2f mesos %.2f", server.Info.Rates.Exp, server.Info.Rates.Drop, server.Info.Rates.Mesos)))
+	hRates := hex.EncodeToString(h.Sum(nil))
+
+	if hRates != server.Info.DefaultRatesChecksum { // Rates event
+		server.Info.Ribbon = 1
```
Hucaru, line 216: Have you experimented with this byte in the login server? https://github.com/Hucaru/Valhalla/blob/master/login/packets.go#L203

channel/player.go
amityahav: If I remember correctly this is an expected behaviour from maple story

channel/handlers.go
amityahav: I thought that passing floats in a packet is complicated, didn't dig in enough probably

world/handlers.go
amityahav: Changed it to 1 and nothing happend, will try different values

amityahav:
CR followup

LuneArk:
If the changes are okay with @Hucaru, what would you be interested in working on next, @amityahav?

world/handlers.go
```diff
@@ -176,6 +186,38 @@ func (server *Server) handlePartyEvent(conn mnet.Server, reader mpacket.Reader)
...
+		server.Info.Rates.Drop = rate
+	case 3:
+		server.Info.Rates.Mesos = rate
+	}
+
+	h := sha512.New()
+	h.Write([]byte(fmt.Sprintf("exp %.2f drop %.2f mesos %.2f", server.Info.Rates.Exp, server.Info.Rates.Drop, server.Info.Rates.Mesos)))
+	hRates := hex.EncodeToString(h.Sum(nil))
+
+	if hRates != server.Info.DefaultRatesChecksum { // Rates event
```
Hucaru, line 210: Why are the values being hashed instead of comparing via an if statement? The hash will be slower. Something like:
```
if server.Info.Rates.DefaultExp != server.Info.Rates.Exp || 
    server.Info.Rates.DefaultMesos != server.Info.Rates.Mesos || ... {
}
``

**Comment by Hucaru** [comment]
Changes look ok to me. Do you have any further changes you want to make? If not I will merge this in

> If the changes are okay with @Hucaru, what would you be interested in working on next, @amityahav?

I think issue #55 might be a good one to do as after that only preventing a character from being created on a world is left as a login world feature.

File: world/handlers.go (hunk 11 above)
Comment (amityahav) [reply world/handlers.go:210]: it seemed like a nasty if statement that i wanted to avoid, but i guess hashing is way more computionally expensive

**Comment by LuneArk** [comment]
> I think issue #55 might be a good one to do as after that only preventing a character from being created on a world is left as a login feature.

Could you double check if the readme is up to date?

@ErwinsExpertise did a lot of work, but I'm unsure if `Player use item (scrolls, potions etc)` and `Player drop item(s)` were completely implemented in [#51](https://github.com/Hucaru/Valhalla/pull/51). There were a lot of stuff on ErwinsExpertise's plate, and [`Player skill logic (haste etc)`](https://github.com/Hucaru/Valhalla/pull/51#issuecomment-1170386147) was also among them. Not sure how that's going right now, but I hope we hear back eventually.

**Comment by Hucaru** [comment]
I think the readme is missing 3 items for the login server
```
[ ] Show EULA on first login
[ ] Show gender select on first login
[ ] Allow world locking when full 
```
I think the last one is managed by a single byte in the world description packet

amityahav:
CR follow up

Hucaru (approved):

Hucaru:
Happy for me to merge?

amityahav:
Yes! Thanks

PR #69, opened 2025-08-15
Add Scrolling, town scrolls, helm chart, and optimizations

ErwinsExpertise:
Sorry for the massive PR, I've started working on this again and forgot to push upstream. 

This PR adds in scrolling for items, return/town scrolls, and a few optimizations.

Additionally, I prefer to use K8S so I've put together a Helm chart as well. Please let me know if anything is needed. I'm going to continue working on this from my end. I'm going to try to *finally* get buff skills working. :)

### DIFF
--- a/channel/handlers.go
+++ b/channel/handlers.go
@@ -913,8 +913,8 @@ func (server *Server) playerUseScroll(conn mnet.Client, reader mpacket.Reader) {
 	}
 
 	// Load items
-	scroll := findUseItemBySlot(plr, scrollSlot)
-	equip := findEquipBySlot(plr, targetSlot)
+	scroll := plr.findUseItemBySlot(scrollSlot)
+	equip := plr.findEquipBySlot(targetSlot)
 	if scroll == nil || equip == nil || scroll.amount < 1 || equip.amount != 1 {
 		plr.send(packetPlayerNoChange())
 		return
@@ -951,8 +951,8 @@ func (server *Server) playerUseScroll(conn mnet.Client, reader mpacket.Reader) {
 
 	if successRoll < int(scrollMeta.Success) {
 		// Success: apply stats, decrement slot, increment scroll count
-		applyScrollEffects(equip, scrollMeta)
-		incrementScrollCount(equip)
+		equip.applyScrollEffects(scrollMeta)
+		equip.incrementScrollCount()
 
 		// Persist and update in-memory slice
 		equip.save(plr.id)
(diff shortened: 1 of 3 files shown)


channel/player.go
```diff
@@ -154,14 +161,70 @@ type player struct {
...
+
+func (d *player) randIntn(n int) int {
+	d.ensureRNG()
+	return d.rng.Intn(n)
+}
+
+// levelUpGains returns (hpGain, mpGain) using per-player RNG and job family.
+// The random component uses a small range similar to legacy behavior.
+// Tweak the constants to match your balance targets if needed.
+func (d *player) levelUpGains() (int16, int16) {
```
Hucaru, line 200: FYI: I believe there is an int scalling component to MP gains which is what fueled "HP washing"

channel/player.go
```diff
@@ -1571,13 +1628,42 @@ func packetInventoryRemoveItem(item item) mpacket.Packet {
 	return p
 }
 
-func packetInventoryChangeEquip(char player) mpacket.Packet {
+func packetInventoryChangeEquip(chr player) mpacket.Packet {
 	p := mpacket.CreateWithOpcode(opcode.SendChannelPlayerChangeAvatar)
-	p.WriteInt32(char.id)
-	p.WriteByte(1)
-	p.WriteBytes(char.displayBytes())
```
Hucaru, line 1578: Why remove the use displayBytes() is there some difference between the standard display bytes and what is needed here?

channel/monster.go
```diff
@@ -80,7 +82,7 @@ func createMonsterFromData(spawnID int32, life nx.Life, m nx.Mob, dropsItems, dr
 		exp:           int32(m.Exp),
 		revives:       m.Revives,
 		summonType:    -2,
-		boss:          m.Boss >= 0,
```
Hucaru, line 83: 😅

channel/handlers.go
```diff
@@ -762,6 +799,335 @@ func (server Server) playerUseInventoryItem(conn mnet.Client, reader mpacket.Rea
...
+		} else {
+			// Normal fail (slot consumed): persist and send full update too
+			equip.save(plr.id)
+			plr.updateItem(*equip)
+			plr.send(packetInventoryAddItem(*equip, true))
+			plr.send(packetUseScroll(plr.id, false, false, false))
+		}
+	}
+}
+
```
Hucaru, line 981: I think the following helper functions should be moved into the items.go file

channel/handlers.go
```diff
@@ -762,6 +799,335 @@ func (server Server) playerUseInventoryItem(conn mnet.Client, reader mpacket.Rea
...
+			return
+		}
+	} else {
+		// Target not found: send failure (TODO: precise packet)
+		plr.send(packetPlayerNoChange())
+	}
+}
+
+// applyBuffItemToPlayer is a placeholder for your buff item usage on a target player.
+func applyBuffItemToPlayer(target *player, itemID int32) {
```
Hucaru: This could be re-written as `func(p* player) applyBuffItem(itemID int32)` to align with the rest of the codebase.

Hucaru (commented):
Thanks for this PR! I have not bothered to review the deployment changes. I have left a few comments but they are all pretty minor things.

channel/handlers.go
```diff
@@ -762,6 +799,335 @@ func (server Server) playerUseInventoryItem(conn mnet.Client, reader mpacket.Rea
...
+	return (itemID % 10000) - (itemID % 100)
+}
+
+// validateScrollTarget performs basic compatibility checks between the scroll and target equip.
+func validateScrollTarget(scrollID int32, equipID int32) bool {
+	return itemTypeToScrollType(equipID) == getScrollType(scrollID)
+}
+
+// applyScrollEffects mutates the equip with the scroll increments from NX.
+func applyScrollEffects(equip *item, scroll nx.Item) {
```
Hucaru: This should be changed into an item method to keep in line with the rest of the codebase

channel/handlers.go
```diff
@@ -762,6 +799,335 @@ func (server Server) playerUseInventoryItem(conn mnet.Client, reader mpacket.Rea
...
+	equip.matk += int16(scroll.IncMAD)
+	equip.mdef += int16(scroll.IncMDD)
+	equip.accuracy += int16(scroll.IncACC)
+	equip.avoid += int16(scroll.IncEVA)
+
+	equip.speed += int16(scroll.IncSpeed)
+	equip.jump += int16(scroll.IncJump)
+}
+
+func incrementScrollCount(equip *item) {
```
Hucaru: Same as `applyScrollEffects`

channel/handlers.go
```diff
@@ -762,6 +799,335 @@ func (server Server) playerUseInventoryItem(conn mnet.Client, reader mpacket.Rea
...
+	equip.scrollLevel++
+}
+
+// removeEquipAtSlot removes the equip from the given slot (equipped negative or inventory positive).
+func removeEquipAtSlot(plr *player, slot int16) bool {
```
Hucaru: This should be a method on player to keep in line with rest of codebase

channel/handlers.go
```diff
@@ -762,6 +799,335 @@ func (server Server) playerUseInventoryItem(conn mnet.Client, reader mpacket.Rea
...
+			}
+			plr.equip[i].amount = 0
+			return true
+		}
+	}
+	return false
+}
+
+// findUseItemBySlot returns the use item (scroll) at the given slot from USE inventory.
+func findUseItemBySlot(plr *player, slot int16) *item {
```
Hucaru: Same as `removeEquipAtSlot`, could probably be made into a general find with a switch statement for slot type removing the need for the other function

channel/handlers.go
```diff
@@ -762,6 +799,335 @@ func (server Server) playerUseInventoryItem(conn mnet.Client, reader mpacket.Rea
...
+	for i := range plr.use {
+		if plr.use[i].slotID == slot {
+			return &plr.use[i]
+		}
+	}
+	return nil
+}
+
+// findEquipBySlot returns the equip by slot (negative = equipped, positive = inventory slot).
+func findEquipBySlot(plr *player, slot int16) *item {
```
Hucaru: Same as `removeEquipAtSlot`

Hucaru:
Unfortunately I have not looked through that part of the idb so I can't assist. However, I think the player buf system might be a good one to tackle first. 

You will need to implement a player buf tick method that gets called by the field instance update method to handle timeouts etc. A good buf to implement first mgiht be GM hide?

channel/player.go
ErwinsExpertise: Is that something we want to solve for? In a lot of way it was actually a bug that was exploited as it was never intended.

PR #70, opened 2025-08-18
Add Initial Buff Support

ErwinsExpertise:
This will add in the initial support for buff skills, and can easily be extended to items. I plan on increasing the support to items, and more skills but I wanted to add in the basic system.

Right now basic buff skills such as haste, rage, etc are supported. I am working on boost, shadow partner, and things of that nature next.

### DIFF
--- a/channel/handlers.go
+++ b/channel/handlers.go
@@ -2994,7 +2994,7 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
 	}
 }
 
-func (server *Server) getAffectedPartyMembers(p *party, src *player, affected byte) []*player {
+func getAffectedPartyMembers(p *party, src *player, affected byte) []*player {
 	if p == nil || src == nil {
 		return nil
 	}
@@ -3087,7 +3087,7 @@ func (server *Server) playerSpecialSkill(conn mnet.Client, reader mpacket.Reader
 			plr.addBuff(skillID, skillLevel, delay)
 			plr.inst.send(packetPlayerSkillAnimThirdParty(plr.id, false, true, skillID, skillLevel))
 		} else {
-			affected := server.getAffectedPartyMembers(plr.party, plr, partyMask)
+			affected := getAffectedPartyMembers(plr.party, plr, partyMask)
 			for _, member := range affected {
 				if member == nil {
 					continue


channel/player.go
```diff
@@ -162,6 +162,8 @@ type player struct {
 
 	rates *rates
 
+	buffs *CharacterBuffs
```
Hucaru, line 165: Does this need to be a pointer, can't it just be embeded directly?

channel/buffs.go
```diff
@@ -0,0 +1,647 @@
...
+	AddSkillBuff(int32(skill.GMHaste), BuffSpeed, BuffJump)
+	AddSkillBuff(int32(skill.GMHolySymbol), BuffHolySymbol)
+	AddSkillBuff(int32(skill.Hide), BuffInvincible)
+}
+
+func init() {
+	LoadBuffs()
+}
+
+// CharacterBuffs is a Go adaptation of the provided C# CharacterBuffs class.
```
Hucaru: Do you have a reference to the C# codebase? We should add it to the readme acknowledgements section.

channel/player.go
ErwinsExpertise: Ideally we keep it as a pointer. The buffs manager owns a mutex, timers, and maps; copying it by value would copy the mutex and live state, which is unsafe and can cause races or deadlocks. We want a single authoritative buff state per character; a pointer guarantees all code paths mutate the same instance, and pointer also allows nil for lazy init.

channel/buffs.go
ErwinsExpertise: I used WvsGlobal as the reference for it.

channel/player.go
Hucaru: I have given it a quick look through and I think you could do a lock free implementation by taking advantage of https://github.com/Hucaru/Valhalla/blob/master/server_channel.go#L242

Have you given any thought to how to track buffs between channel changes and login in and out?

channel/handlers.go
```diff
@@ -133,6 +134,10 @@ func (server *Server) HandleClientPacket(conn mnet.Client, reader mpacket.Reader
 		server.npcMovement(conn, reader)
 	case opcode.RecvChannelBoatMap:
 		// [mapID int32][? byte]
+	case opcode.RecvChannelAcknowledgeBuff:
+		// Consume
```
Hucaru, line 138: I assume the server needs to do something here?

channel/handlers.go
```diff
@@ -2979,3 +2993,167 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
 		log.Println("Unknown chat event type:", op)
 	}
 }
+
+func (d *player) removeAllCooldowns() {
```
Hucaru: Should live in `player.go`

channel/handlers.go
```diff
@@ -2979,3 +2993,167 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
...
+	if d == nil || d.skills == nil {
+		return
+	}
+	for id, ps := range d.skills {
```
Hucaru: If you want to ignore id value then this can be re-written as `for _, ps := range d.skills {`

channel/handlers.go
```diff
@@ -2979,3 +2993,167 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
...
+		ps.Cooldown = 0
+		ps.TimeLastUsed = 0
+		d.updateSkill(ps)
+		_ = id
+	}
+}
+
+func (server *Server) getAffectedPartyMembers(p *party, affectedMask byte) []*player {
```
Hucaru: Should you not be taking map and instance into account for this? Party members on the same map in the same instance should be affected

channel/handlers.go
```diff
@@ -2979,3 +2993,167 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
...
+		for i := 0; i < count; i++ {
+			_ = reader.ReadInt32() // mob spawn id
+		}
+		_ = reader.ReadInt16() // delay
+	}
+	readXY := func() {
+		_ = reader.ReadInt16()
+		_ = reader.ReadInt16()
+	}
+
```
Hucaru, line 3081: `readXY` and `readMobListAndDelay` don't need to be functions, they can just be inlined since they are not reused

channel/handlers.go
```diff
@@ -2979,3 +2993,167 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
...
+	applied := false
```
Hucaru: This variable is never used?

channel/handlers.go
```diff
@@ -2979,3 +2993,167 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
...
+
+	switch skill.Skill(skillID) {
+	// Party buffs handled earlier remain unchanged...
+	case skill.Haste, skill.BanditHaste, skill.Bless, skill.IronWill, skill.Rage,
+		skill.Meditation, skill.ILMeditation, skill.MesoUp, skill.HolySymbol, skill.HyperBody:
+		if plr.party == nil {
+			plr.addBuff(skillID, skillLevel, delay)
+			plr.send(packetPlayerSkillAnimThirdParty(plr.id, false, true, skillID, skillLevel))
```
Hucaru: This should be sent to the instance as other players should see when you have applied a skill buff and should be applied to all the cases in this switch statement.

channel/handlers.go
```diff
@@ -2979,3 +2993,167 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
...
+
+	// Summons and puppet: [x short][y short]
+	case skill.SummonDragon,
+		skill.SilverHawk, skill.GoldenEagle,
+		skill.Puppet, skill.SniperPuppet:
+		readXY()
+
+	default:
+		// Always send a self animation so client shows casting even for non-buffs.
+		plr.send(packetPlayerSkillAnimThirdParty(plr.id, false, true, skillID, skillLevel))
```
Hucaru: This should also go to instance I suspect

channel/handlers.go
```diff
@@ -2979,3 +2993,167 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
...
+		plr.send(packetPlayerNoChange())
+		return
+	}
+
+	_ = applied
+}
+
+func (server *Server) playerCancelBuff(conn mnet.Client, reader mpacket.Reader) {
+	plr, err := server.players.getFromConn(conn)
+	if err != nil || plr == nil || plr.buffs == nil {
```
Hucaru: Don't think we need to validate nil plr here as we don't elsewhere

channel/player.go
```diff
@@ -1378,6 +1397,143 @@ func packetPlayerLevelUpAnimation(charID int32) mpacket.Packet {
...
+func packetPlayerSkillAnimSelf(charID int32, skillID int32, level byte) mpacket.Packet {
+	p := mpacket.CreateWithOpcode(opcode.SendChannelPlayerAnimation)
+	p.WriteInt32(charID)
+	p.WriteByte(0x01)
+	p.WriteInt32(skillID)
+	p.WriteByte(level)
+	return p
+}
+
+func packetPlayerSkillAnimThirdParty(charID int32, party bool, self bool, skillID int32, level byte) mpacket.Packet {
```
Hucaru, line 1496: What does "ThirdParty" mean in this context?

channel/player.go
```diff
@@ -1378,6 +1397,143 @@ func packetPlayerLevelUpAnimation(charID int32) mpacket.Packet {
...
+		p.WriteByte(extra)
+	}
+
+	p.WriteInt64(0)
+	p.WriteInt64(0)
+
+	return p
+}
+
+func packetPlayerGiveForeignBuff(charID int32, mask []byte, values []byte, extra byte) mpacket.Packet {
```
Hucaru, line 1550: Is this applying a buff to other players e.g. when applying haste this packet is sent for each party member sharing the buff?

channel/player.go
```diff
@@ -1378,6 +1397,143 @@ func packetPlayerLevelUpAnimation(charID int32) mpacket.Packet {
...
+
+func packetPlayerCancelForeignBuff(charID int32, mask []byte) mpacket.Packet {
+	p := mpacket.CreateWithOpcode(opcode.SendChannelPlayerDebuff)
+	p.WriteInt32(charID)
+	p.WriteBytes(mask)
+	p.WriteUint64(0)
+	return p
+}
+
+func packetPlayerShowBuffEffect(charID int32, skillID int32, effectID int32) mpacket.Packet {
```
Hucaru, line 1615: This is what is sent to other players excl. party members affected?

channel/handlers.go
ErwinsExpertise: None of the other sources I looked at actually do anything with this. I'm assuming it's just an anti-cheat, or sync mechanism?

channel/handlers.go
ErwinsExpertise: Ah, sorry debug value I forgot to remove.

channel/handlers.go
ErwinsExpertise: I kept them as they are so that if anyone else picks up they know which functions are needed for each skill. It can be removed/commented otherwise, but this was just a quick way.

channel/player.go
ErwinsExpertise: It's the packet used to apply the animation to another character. As an example when you used haste and the animation appears above the character.

channel/player.go
ErwinsExpertise: Yes.

channel/player.go
ErwinsExpertise: I'm not actually sure what this packet is for? I found it in other sources, and when looking through the IDB, but I couldn't make out it's actual use at the time. I put it in just so it's there in-case it is needed in the future.

channel/handlers.go
ErwinsExpertise: Added logic for taking into account the party mask and map

channel/server.go
```diff
@@ -292,6 +295,85 @@ func (server *Server) ClientDisconnected(conn mnet.Client) {
 	}
 }
 
+func (server *Server) saveBuffSnapshot(plr *player) {
```
Hucaru: These functions feel like they should be methods of the player struct rather than the server one

channel/item.go
```diff
@@ -461,9 +463,14 @@ func (v item) use(plr *player) {
 		plr.giveMP(v.mp)
 	}
 
-	// Need to add stat buffs (W.ATT, M.ATT, etc)
-	// This will require timers to ensure buffs are removed once finished
+	if plr == nil {
+		return
```
Hucaru: This will never reach as line 455 will cause a panic if plr is nil

mpacket/packet.go
```diff
@@ -205,3 +210,35 @@ func (p *Packet) readFloat32(pos *int) float32 {
 func (p *Packet) readString(pos *int, length int) string {
 	return string(p.readBytes(pos, length))
 }
+
+func (p *Packet) Position() int {
```
Hucaru, line 214: Where are the following functions used? I can't see where they are called in this PR.

ErwinsExpertise: Just a helper method I initially added to get position for mask shifting, but it's not really needed. I forgot about it to be honest.

channel/flags.go
```diff
@@ -0,0 +1,451 @@
...
+
+import (
+	"encoding/hex"
+	"fmt"
+	"math/rand"
+	"strings"
+	"time"
+)
+
+type Flag struct {
```
Hucaru: I don't think this needs to be a struct, you can just typedef []uint32 and attach methods onto the type e.g.:
```
type Flag []uint32
func (f *Flag) CompareTo(other *Flag) int {
...
```

channel/flags.go
```diff
@@ -0,0 +1,451 @@
...
+	if f.aData[uLen] < value {
+		return -1
+	}
+	return 0
+}
+
+// DecodeBuffer fills the flag from a provided 32-bit decoder function.
+// The decoder should return the next 4-byte unsigned value in the
+// same order as the original wire format.
+func (f *Flag) DecodeBuffer(decode4 func() uint32) {
```
Hucaru: Is this used anywhere? ctrl+f can't find any isntances of it's use.

channel/buffs.go
```diff
@@ -0,0 +1,744 @@
...
+
+func (cb *CharacterBuffs) scheduleExpiryLocked(skillID int32, after time.Duration) {
+	// Cancel previous
+	if t, ok := cb.expireTimers[skillID]; ok && t != nil {
+		t.Stop()
+		delete(cb.expireTimers, skillID)
+	}
+
+	if after <= 0 {
+		cb.post(func() { cb.expireBuffNow(skillID) })
```
Hucaru, line 595: It would be safer to instead pass the playerID and skillID and perform a player lookup as part of the post job that way you don't have to be concerned about stale pointers.

channel/handlers.go
```diff
@@ -3044,8 +3058,8 @@ func (server *Server) playerSpecialSkill(conn mnet.Client, reader mpacket.Reader
 		return
 	}
 
-	_ = reader.ReadByte()       // party flags
-	delay := reader.ReadInt16() // delay
+	partyMask := reader.ReadByte() // party flags
```
Hucaru, line 3067: What does this mask contain?

channel/server.go
ErwinsExpertise: Agreed, I moved it.

channel/item.go
ErwinsExpertise: Removed

channel/flags.go
ErwinsExpertise: True, I had as a struct for future flexibility, but it can be a typedef too. Changed as requested.

channel/buffs.go
ErwinsExpertise: This is already checked in the expireBuffNow func.

channel/handlers.go
ErwinsExpertise: It's which party members are affected in AoE. I changed the code a little bit. It just shows which characters in party are affected and in range based on id.

Hucaru: Ah I didn't realise the client provided this

ErwinsExpertise: Yeah, I kept map checks and such in-place just in-case, but they're not technically necessary.

channel/flags.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	words := uBits >> 5 // divide by 32
+	f := Flag(make([]uint32, words))
+	f.SetValue(0)
+	return &f
+}
+
+// Data returns the internal data slice (do not modify).
+func (f *Flag) Data() []uint32 {
+	if f == nil {
```
Hucaru: As a general rule I don't think pointer receivers should be checked for nil. It should be left to panic as it's a major bug if we end up in this state

channel/handlers.go
```diff
@@ -2994,33 +2994,49 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
 	}
 }
 
-func (server *Server) getAffectedPartyMembers(p *party, src *player, includeSelf bool) []*player {
+func (server *Server) getAffectedPartyMembers(p *party, src *player, affected byte) []*player {
```
Hucaru: I think this can just be a function not a method as I don't see where this function reads/writes to server state?

Hucaru:
@ErwinsExpertise I think this is at a stage that we can merge in with further refinement over time as the codebase matures. Are you happy for me to merge?

Hucaru (approved):

PR #75, opened 2025-08-20
Add NPC Scripts

ErwinsExpertise:
This PR will add a ton of additional NPC scripts to the source. Rather than updated by hand I found a tool to run it through an llm and convert them. 

They were all converted using [this converter](https://github.com/lastbattle/gpt_stateless_npc_converter/tree/main) and it was hooked up to a full size 'kimi-k2-instruct' model to ensure the best possible outcome. I've tested *most* of the scripts and they all seem to work. There might be some with functions that need to be implemented, but I've looked and implemented any that weren't already. 

You can see the prompt used on my fork here:
https://github.com/ErwinsExpertise/gpt_stateless_npc_converter/blob/main/prompts/2_system.txt

### DIFF
--- a/channel/script.go
+++ b/channel/script.go
@@ -191,6 +191,36 @@ func (tracker *npcChatStateTracker) popState() {
 	tracker.lastPos--
 }
 
+func (tracker *npcChatStateTracker) currentSelection() int32 {
+	if tracker.selection >= 0 && tracker.selection < len(tracker.selections) {
+		return tracker.selections[tracker.selection]
+	}
+	if len(tracker.selections) > 0 {
+		return tracker.selections[len(tracker.selections)-1]
+	}
+	return 0
+}
+
+func (tracker *npcChatStateTracker) currentInputString() string {
+	if tracker.input >= 0 && tracker.input < len(tracker.inputs) {
+		return tracker.inputs[tracker.input]
+	}
+	if len(tracker.inputs) > 0 {
+		return tracker.inputs[len(tracker.inputs)-1]
+	}
+	return ""
+}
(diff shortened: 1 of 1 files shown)


Hucaru:
This is one of the times where using an LLM is actually a good choice!

I am going to asume the LLM is correct and am not going to check the scripts. Have you done any testing on a sample of these scripts?

Are you ok with leaving this PR up and having it go in after I merge in my guild PR? I only have a few things left to implement and then merge in master and resolve conflicts.

Hucaru:
> > This is one of the times where using an LLM is actually a good choice!
> > I am going to asume the LLM is correct and am not going to check the scripts. Have you done any testing on a sample of these scripts?
> > Are you ok with leaving this PR up and having it go in after I merge in my guild PR? I only have a few things left to implement and then merge in master and resolve conflicts.
> 
> That sounds great to me!
> 
> Most of the scripts are correct, and if there were missing functions it was instructed to make a script as such:
> 
> ```
> npc.sendOk("NPC ID is not yet implemented.")
> ```
> 
> I used another script to check the NPC id's from NX and remove any NPC's that don't exist in the wz file to ensure there aren't any extras. I'm going to keep going through and finding functions I might have missed and implement them. I'm also going to start experimenting with quests.

I started looking at quests before I got sidelined a long time ago but you can see how the client handles quests and their state from looking at the player login packet https://github.com/Hucaru/Valhalla/blob/master/channel/player.go#L1813. I think reactors are also a good one to look into at the same time as some only spawn if a quest is active I believe (it was a few years ago I looked into them).

PR #76, opened 2025-08-20
Write‑behind persistence for character state + reliable disconnect saves

ErwinsExpertise:
This PR makes character saving more reliable and a lot lighter on the database. Instead of writing to the DB on every stat change, we mark only the fields that changed and let a background saver coalesce those updates over a short window. That means AP/SP clicks, mesos adjustments, and other frequent changes get merged and written once, touching only the relevant columns. On disconnect (graceful or not), we now look up the player for that socket, flush any pending changes, and do a final checkpoint save so progress isn’t lost. There’s also a periodic autosave and a shutdown checkpoint to bound worst‑case loss even further. Skill upserts have been fixed so level and cooldown actually update, and quick updates consistently target rows by primary key.
Under the hood, this separates “gameplay state” from “durability.” The player object remains focused on game logic, while a small write‑behind component owns when and how data hits the DB. That separation lets us debounce, batch, and write only what changed, which scales far better than writing directly from the main player struct on every adjustment. It also makes lifecycle moments—disconnects, autosaves, shutdown—easy places to flush safely without hammering the database.


channel/session.go
```diff
@@ -0,0 +1,53 @@
...
+	plr := detachSession(conn)
+	if plr == nil {
+		return
+	}
+	defer func() {
+		if r := recover(); r != nil {
+			log.Printf("panic during OnDisconnect logout for player %d: %v", plr.id, r)
+		}
+	}()
+	plr.Logout("disconnect")
```
Hucaru: I think this can be done in the main thread. If `plr.Logout` is changed from a pointer receiver to a copy receiver then it is thread safe and you can invoke it via `go plr.logout("disconnect)`. This way the session registry won't be needed

channel/player.go
```diff
@@ -1059,6 +1081,27 @@ func (d player) displayBytes() []byte {
 	return pkt
 }
 
+// Logout flushes coalesced state and does a full checkpoint save.
+func (d *player) Logout(reason string) {
+	log.Printf("player(%d) logout: %s", d.id, reason)
```
Hucaru: I think we should avoid writing to stdout as much as possible (preferebly just errors)

channel/saver.go
```diff
@@ -0,0 +1,329 @@
...
+	DirtyEXP
+	DirtyMap
+	DirtyJob
+	DirtyLevel
+	DirtyStr
+	DirtyDex
+	DirtyInt
+	DirtyLuk
+	DirtyFame
+)
```
Hucaru, line 35: I think you want to also add the following fields to snapshot
```
	equipSlotSize byte
	useSlotSize   byte
	setupSlotSize byte
	etcSlotSize   byte
	cashSlotSize  byte

	miniGameWins, miniGameDraw, miniGameLoss, miniGamePoints int32
	buddyListSize byte
```

You could probably ignore slot size as I imagine they won't change that often but maybe minigames might?

channel/saver.go
```diff
@@ -0,0 +1,329 @@
...
+// pendingSave is a coalesced, debounced update for one character.
+type pendingSave struct {
+	due  time.Time
+	bits DirtyBits
+	snap snapshot
+}
+
+// Saver batches and persists character updates.
+type Saver struct {
+	mu      sync.Mutex
```
Hucaru: I think this should take advantage of the job submit pattern instead of using mutex with the `persist` function call taking a copy and being executed via `go persist`

channel/saver.go
```diff
@@ -0,0 +1,329 @@
...
+	pending map[int32]*pendingSave // charID -> pendingSave
+	ticker  *time.Ticker
+	stopCh  chan struct{}
+}
+
+// SaverInstance is the global saver used by the channel server.
+var SaverInstance *Saver
+
+// StartSaver boots the saver loop; should be called once at startup.
+func StartSaver() {
```
Hucaru: I think changing this to a package `init()` function would be more idiomatic

channel/saver.go
```diff
@@ -0,0 +1,329 @@
...
+// Saver batches and persists character updates.
+type Saver struct {
+	mu      sync.Mutex
+	pending map[int32]*pendingSave // charID -> pendingSave
+	ticker  *time.Ticker
+	stopCh  chan struct{}
+}
+
+// SaverInstance is the global saver used by the channel server.
+var SaverInstance *Saver
```
Hucaru: I think a better singleton pattern would be not to export this variable or the type and use functions to perform the work

channel/server.go
```diff
@@ -302,3 +300,35 @@ func (server *Server) ClientDisconnected(conn mnet.Client) {
...
+	flushTick := time.NewTicker(30 * time.Second)
+
+	go func() {
+		defer flushTick.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-flushTick.C:
+				for _, p := range server.players {
```
Hucaru: I don't think this access is thread safe. To make it thread safe the go function should submit the flush iteration back as a job onto the main thread

ErwinsExpertise: @Hucaru pushed some changes to do this. Let me know what you think

Hucaru (approved):

channel/server.go
Hucaru: The implementation is fine. Another way of doing the same thing would be to use a ticker e.g. 
```
go func() {
    ticker := time.NewTicker(time)
    select {
        case t := <-ticker.C:
            // submit job to dispatch
    }
}

```

PR #77, opened 2025-08-21
Quest system

ErwinsExpertise:
This PR will add pretty much full support for quest systems. 99% of the actual handling is done client side, but there's still a few things we have to do like keep track of in-progress/completed quests. We also do requirement checks, and the reward system.


channel/quests.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	out := make([]quest, 0, len(q.completed))
+	for _, v := range q.completed {
+		out = append(out, v)
+	}
+	return out
+}
+
+// Runtime update packets
+func packetCompleteQuest(questID int16) mpacket.Packet {
```
Hucaru: It would be preferable if the function names where `packetQuest....` makes it easier when searching for quest related packets and if a developer is using code completion

channel/quests.go
```diff
@@ -0,0 +1,256 @@
...
+import (
+	"time"
+
+	"github.com/Hucaru/Valhalla/common"
+	"github.com/Hucaru/Valhalla/common/opcode"
+	"github.com/Hucaru/Valhalla/mpacket"
+)
+
+const (
+	QUEST_LOST_ITEM = 0x00
```
Hucaru: Constants are lower case (unexported) and use camel case like rest of codebase

channel/pools.go
```diff
@@ -428,6 +447,16 @@ func rollDrop(r *rand.Rand, baseChance int64, rate float32) bool {
 	return r.Int63n(denom) < scaled
 }
 
+func (d *player) allowsQuestDrop(qid int32) bool {
```
Hucaru: Function should probably live in `player.go`

channel/player.go
```diff
@@ -1398,36 +1419,313 @@ func (d *player) loadAndApplyBuffSnapshot() {
...
 		log.Println("loadBuffSnapshot rows:", err)
 		return
 	}
 
+	if len(toDelete) > 0 {
+		var b strings.Builder
+		b.WriteString("DELETE FROM character_buffs WHERE characterID=? AND sourceID IN (")
+		for i := range toDelete {
+			if i > 0 {
+				b.WriteByte(',')
```
Hucaru, line 1460: Would prefer hex codes rather than ascii unless the ascii representation contains meaning?

channel/handlers.go
```diff
@@ -779,7 +800,24 @@ func (server Server) warpPlayer(plr *player, dstField *field, dstPortal portal)
 	// plr.mapPos = dstPortal.id
 	plr.pos = dstPortal.pos
 
-	plr.send(packetMapChange(dstField.id, int32(server.id), dstPortal.id, plr.hp)) // plr.ChangeMap(dstField.ID, dstPortal.ID(), dstPortal.Pos(), foothold)
+	spawnIdx, idxErr := dstInst.calculateNearestSpawnPortalID(dstPortal.pos)
+	if idxErr != nil {
+		spawnIdx = dstPortal.id
+	}
+
+	packetMapChange := func(mapID int32, channelID int32, mapPos byte, hp int16) mpacket.Packet {
```
Hucaru: Don't we already have this packet defined elsewhere? If not then it should be added to this file as a function

channel/quests.go
ErwinsExpertise: Sorry about that! I actually meant to put them all in the internal package but forgot.

internal/codes.go
Hucaru, line 1: I would place these constants in https://github.com/Hucaru/Valhalla/blob/master/constant/constants.go#L103 the internal opcodes are the interserver communication not client

ErwinsExpertise: @Hucaru Ah okay, I moved them.

PR #79, opened 2025-08-23
Feature/guild system

Hucaru:
- Added guild system #40 
- Improved party system cross channel. Slight bug where channel might not send party member party packet on login.

guild_1
guild_2
guild_3


channel/commands.go
```diff
@@ -763,14 +848,73 @@ func (server *Server) gmCommand(conn mnet.Client, msg string) {
...
+			conn.Send(packetMessageRedText(err.Error()))
+			return
+		}
+
+		if plr.guild != nil {
+			conn.Send(packetMessageRedText("Already in a guild"))
+			return
+		}
+
+		_ = guildName
```
ErwinsExpertise: Why is this inline? Is this supposed to actually create a guild?

channel/handlers.go
```diff
@@ -2593,14 +2669,317 @@ func (server Server) roomWindow(conn mnet.Client, reader mpacket.Reader) {
...
 	default:
 		log.Println("Unknown room operation", operation)
 	}
 }
 
+func (server *Server) guildManagement(conn mnet.Client, reader mpacket.Reader) {
+	op := reader.ReadByte()
+
+	switch op {
+	case 0x02: // create guild name dialogue
```
ErwinsExpertise: It would be a lot cleaner, and easier to read if these opcodes were constants in a central place.

channel/commands.go
Hucaru: Ah good catch, this whole command is useless as is. I will change it to just start the guild npc

channel/handlers.go
Hucaru: Done

Issue #85, opened 2025-08-25
Server does not gracefully shutdown


Hucaru:
@ErwinsExpertise the channel server hangs when receiving shutdown signal. I think this needs another look

PR #86, opened 2025-08-25
Resolve deadlocks

ErwinsExpertise:
I observed some deadlocks happening after the snapshot logic was merged in. This PR fixes two classes of deadlocks by making the main-thread dispatcher the single source of truth. The dispatcher now signals when it’s ready and we only start things that enqueue work (world registration, autosave, countdown messages) after that point. Anything running inside the dispatcher calls handlers directly instead of posting back to itself, while background goroutines enqueue into the dispatcher instead of touching state. Disconnects are now authoritative and always clear in-memory state and DB session flags, so “ghost online” characters don’t block reconnects. Also resolved some templating issues with the helm chart.

### DIFF
--- a/helm/templates/configmaps.yaml
+++ b/helm/templates/configmaps.yaml
@@ -48,28 +48,33 @@ data:
     listenAddress = "0.0.0.0"
     listenPort = "8584"
     packetQueueSize = {{ .Values.world.packetQueueSize }}
+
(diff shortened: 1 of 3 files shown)


happytreees:
I think these were likely the cause of #85 @Hucaru

Hucaru (approved):

PR #89, opened 2025-08-25
Feature/Working Summons

ErwinsExpertise:
This PR will add working summons to the source. Everything works

### DIFF
--- a/channel/handlers.go
+++ b/channel/handlers.go
@@ -818,10 +818,10 @@ func (server Server) warpPlayer(plr *player, dstField *field, dstPortal portal) 
 	if plr.summons != nil {
 		if plr.summons.puppet != nil {
 			// Debuff puppet
-			plr.removeSummon(true, 0x01)
+			plr.removeSummon(true, constant.SummonRemoveReasonCancel)
 		}
 		if plr.summons.summon != nil {
-			srcInst.send(packetRemoveSummon(plr.id, plr.summons.summon.SkillID, 0x01))
+			srcInst.send(packetRemoveSummon(plr.id, plr.summons.summon.SkillID, constant.SummonRemoveReasonCancel))
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


channel/summons.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+type summonState struct {
+	puppet  *Summon
+	summon  *Summon
+	puTimer *time.Timer
+	smTimer *time.Timer
+}
+
+func (p *player) ensureSummonState() {
```
Hucaru: This should be in `player.go` or change the function to take player as a standard parameter

channel/summons.go
```diff
@@ -0,0 +1,203 @@
...
+
+import (
+	"time"
+
+	"github.com/Hucaru/Valhalla/common/opcode"
+	"github.com/Hucaru/Valhalla/mpacket"
+)
+
+// Summon represents a player-owned summon/puppet instance.
+type Summon struct {
```
Hucaru: What is the reason for exporting the struct and it's fields outside of the channel package?

channel/summons.go
```diff
@@ -0,0 +1,203 @@
...
+		if p.summons.summon != nil {
+			s.removeSummon(p, false, 0x04) // replaced
+		}
+		p.summons.summon = su
+
+		if p.summons.smTimer != nil {
+			p.summons.smTimer.Stop()
+		}
+		p.summons.smTimer = time.AfterFunc(time.Duration(durationSec)*time.Second, func() {
+			s.removeSummon(p, false, 0x02) // expired
```
Hucaru: What happens if player logs out before this is called?

channel/summons.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func (p *player) ensureSummonState() {
+	if p.summons == nil {
+		p.summons = &summonState{}
+	}
+}
+
+// addSummon registers and broadcasts a summon; replaces any existing same-type summon.
+func (s *Server) addSummon(p *player, su *Summon, durationSec int) {
```
Hucaru: I don't think the server needs to keep track of the summons. They should be attached to the player. When summon information needs to be broadcast at most it's only ever to the player's instance.

channel/handlers.go
```diff
@@ -773,47 +799,59 @@ func (server Server) playerUsePortal(conn mnet.Client, reader mpacket.Reader) {
...
 		if dstInst, err = dstField.getInstance(0); err != nil {
 			return err
 		}
 	}
 
-	err = srcInst.removePlayer(plr)
+	if plr.summons != nil {
+		if plr.summons.puppet != nil {
+			// Debuff puppet
+			plr.removeSummon(true, 0x01)
```
Hucaru: In a few places we have magic numbers that we should clean up

Hucaru:
The refactor is much easier to follow than the initial implementation, apart from the magic numbers I think this is good to merge.

PR #96, opened 2025-09-04
Implement cash shop

ErwinsExpertise:
This PR will implement basic working function of the CS. You can enter/exit, and purchase items. NX/maplepoints are stored in the account table. I added a bunch of extra packets that are "nice-to-haves" like the wishlist, but otherwise unnecessary for basic function. I couldn't really figure out the last couple sections of the set CS packet. It's where you set the best items but I could not get it to work properly. Current implementation is good enough to not crash, but will need additional work to make it work properly. 

Capture

IDB Notes:
CashShop_OnPacket = `sub_44539F`
CCashShop = `sub_43A710`
OnSetCashShop = `sub_56777B`

### DIFF
--- a/channel/player.go
+++ b/channel/player.go
@@ -1058,37 +1058,55 @@ func (d *Player) updateSkill(updatedSkill playerSkill) {
 func (d *Player) useSkill(id int32, level byte, projectileID int32) error {
 	skillInfo, _ := nx.GetPlayerSkill(id)
 
-	if skillUsed, ok := d.skills[id]; ok {
-		if skillUsed.Level != level {
-			d.Conn.Send(packetMessageRedText("skill level mismatch"))
-			return errors.New("skill level mismatch")
-		}
-		if skillInfo[skillUsed.Level].MpCon > 0 {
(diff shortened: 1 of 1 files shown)


channel/handlers.go
```diff
@@ -668,9 +670,47 @@ func validateSkillWithJob(jobID int16, baseSkillID int32) bool {
...
 }
 
-func (server Server) playerUsePortal(conn mnet.Client, reader mpacket.Reader) {
+func (server Server) playerEnterCashShop(conn mnet.Client, reader mpacket.Reader) {
 	plr, err := server.players.getFromConn(conn)
+	if err != nil {
+		return
+	}
+
+	if plr.summons != nil {
```
Hucaru: This can probably be refactored into a function called something along the lines of `plr.hideSummons()` and it probably makes sense to have a `plr.showSummons` as I imagine these would be useful elsewhere as well

Hucaru:
Good work on the reversing! Ideally the cash shop should be a separate server. If you don't want to do that in this PR then create a new top level package called cash shop and place all the relevant handlers and packets in there.

Hucaru (approved):

Issue #93, opened 2025-09-04
Some Portals Send Characters to the Wrong Coordinates of the Next Map


LuneArk:
We'll need to double-check that all portals send characters to the correct coordinates of the next map.

There are many examples of this happening on Maple Island.

PR #100, opened 2025-09-14
Add reactors, scripted portal, npc shops, npc's, and bug fixes

ErwinsExpertise:
This PR introduces a ton of changes and all of them great. 

Quick rundown:
- Reactors ( Boss reactors like Pap, and Zak now work )
- Scripted Portals ( Like FM)
- Throwing Star bugs all resolved
- Added all NPC Shops
- Shop actions now all work(Buying, Selling, Recharging)
- Fixed portal spawn to now use portal instead of spawn point when using portals
- Added a ton of NPC's

zak
pap
shop


channel/handlers.go
```diff
@@ -2163,15 +2243,20 @@ func (server *Server) npcChatStart(conn mnet.Client, reader mpacket.Reader) {
...
 	}
 
 	server.npcChat[conn] = controller
 
-	if controller.run() {
-		delete(server.npcChat, conn)
+	// Run the script. If it returns true, chat flow ended.
+	// If a shop was opened (persistShop), keep the controller for shop ops.
+	ended := controller.run()
+	if ended {
```
Hucaru: This can be cleaned up to something like:
```
if ended && !controller.persistShop {
    delete(server.npcChat, conn)
}
```

channel/pools.go
```diff
@@ -967,6 +981,347 @@ func (pool dropPool) HideDrops(plr *Player) {
...
+	if r.reactorTime > 0 {
+		time.AfterFunc(time.Duration(r.reactorTime)*time.Second, func() {
+			r.state = 0
+			r.frameDelay = 0
+			pool.instance.send(packetMapReactorEnterField(r.spawnID, r.templateID, r.state, r.pos.x, r.pos.y, r.faceLeft))
+		})
+	}
+}
+
+func (pool *reactorPool) TriggerHit(spawnID int32, cause byte) {
```
Hucaru: Is there a reason why these functions are exported?

channel/pools.go
```diff
@@ -967,6 +981,347 @@ func (pool dropPool) HideDrops(plr *Player) {
...
+	for _, r := range pool.reactors {
+		r.state = 0
+		r.frameDelay = 0
+		if send {
+			pool.instance.send(packetMapReactorChangeState(r.spawnID, r.state, r.pos.x, r.pos.y, r.frameDelay, r.faceLeft, 0))
+		}
+	}
+}
+
+type reactorTableEntry map[string]interface{}
```
Hucaru, line 1083: Is this data not stored in the NX/Wz?

channel/script.go
```diff
@@ -509,9 +651,145 @@ func (ctrl *npcChatController) SendGuildEmblemEditor() {
 // SendShop packet to Player
 func (ctrl *npcChatController) SendShop(goods [][]int32) {
 	ctrl.goods = goods
+	ctrl.persistShop = true
```
Hucaru: I am surprised you need to do this. I would expect when the client closes the shop it would send a finished npc chat packet. If it does indeed send a end npc chat packet when the shop window closes I would change this to a vm interrupt then the if check in `handlers.go` can be reverted to what it was before.

channel/script.go
```diff
@@ -521,23 +799,26 @@ func (ctrl *npcChatController) clearUserInput() {
 
 // Selection value
 func (ctrl *npcChatController) Selection() int32 {
-	val := ctrl.stateTracker.selections[ctrl.stateTracker.selection]
-	ctrl.stateTracker.selection++
-	return val
+	if len(ctrl.stateTracker.selections) == 0 {
```
Hucaru, line 812: What's the reason for changing how npc state is tracked?

reactors.json
Hucaru, line 1: Where does this come from? Is it not already in NX/Wz data?

scripts/npc-quests.txt
Hucaru, line 1: Is this a reference document of some kind? If so I think it should probably be formated as an `.md` with a simple explanation at the top if one is needed

scripts/portals/market00.js
Hucaru, line 1: Is FM the only case of a scripted portal in this version of the game? If so I am inclined to just hardcode it's behaviour into the handler directly.

channel/script.go
```diff
@@ -399,10 +526,11 @@ type npcChatController struct {
 
 func createNpcChatController(npcID int32, conn mnet.Client, program *goja.Program, plr *Player, fields map[int32]*field, warpFunc warpFn, worldConn mnet.Server) (*npcChatController, error) {
```
Hucaru, line 544: Have you verified each of the unique behaviours added/changed in the controller works as inteded?

Hucaru:
Does this PR fix: https://github.com/Hucaru/Valhalla/issues/93?

channel/pools.go
ErwinsExpertise: It is not. This is all server-side, because it controls the actions such as spawning Zakum when you drop eye of fire.

channel/script.go
ErwinsExpertise: I don't recall seeing any finish NPC chat packets coming in. I believe all of that is client-sided, however, the server needs to keep the context open for shops, otherwise it's not able to identify which shop is being utilized and doesn't have the proper data. We added in a variable so we only introduce the behavior when we want, otherwise it behaves as normal.

reactors.json
ErwinsExpertise: It is not. This is all server-side, because it controls the actions such as spawning Zakum when you drop eye of fire.

scripts/npc-quests.txt
ErwinsExpertise: Can do, it's just a dump of all of the quests for each NPC with quest-ID and name. Just to be used for reference later if needed for NPC scripting.

scripts/portals/market00.js
ErwinsExpertise: It is not, but that's why I implemented it lol. I believe all PQ's use scripted portals as well. Also boss portals like zak and pap

channel/script.go
ErwinsExpertise: I think there might've been some I didn't test, but I tested almost all. I'll just go through them all tomorrow and verify.

channel/script.go
Hucaru: I don't see where `persistShop` is set to false. If the client doesn't send an end packet this will lead to the controller not being cleaned up and leaking?

ErwinsExpertise: Looks like there is one, didn't see it before:
https://github.com/ErwinsExpertise/Valhalla/blob/master/channel/handlers.go#L2541

ErwinsExpertise: refactored to be an interrupt instead. Seems to work as expected.

Hucaru:
From inspecting the code I think it's down to this:

SendSelection doesn't handle return state:
```
func (ctrl *npcChatController) SendSelection(msg string) {
	if ctrl.stateTracker.performInterrupt() {
		ctrl.conn.Send(packetNpcChatSelection(ctrl.npcID, msg))
		ctrl.vm.Interrupt("SendSelection")
	}
}
```

SendAvatar will return the selection if it's not interrupting:
```
func (ctrl *npcChatController) SendAvatar(text string, avatars ...int32) int {
	if ctrl.stateTracker.performInterrupt() {
		ctrl.stateTracker.addState(npcSelectionState)
		ctrl.conn.Send(packetNpcChatStyleWindow(ctrl.npcID, text, avatars))
		ctrl.vm.Interrupt("SendAvatar")
		return -1
	}

	if len(ctrl.stateTracker.selections) > ctrl.stateTracker.selection {
		val := ctrl.stateTracker.selections[ctrl.stateTracker.selection]
		ctrl.stateTracker.selection++
		return int(val)
	}
	return -1
}
```

The script calls `npc.sendAvatar()` followed by a `npc.selection()`, when I think what you should do with the current code is:

```
switch (npc.sendAvatar()) {
case 0:
case 1:
...
}
```

To be honest I prefere the above and if that is the solution I wouldn't be against converting the `npc.sendSelection()` to work in the same way

channel/script.go
```diff
@@ -666,9 +664,11 @@ func (ctrl *npcChatController) SendGuildEmblemEditor() {
 
 // SendShop packet to Player
 func (ctrl *npcChatController) SendShop(goods [][]int32) {
-	ctrl.goods = goods
-	ctrl.persistShop = true
-	ctrl.conn.Send(packetNpcShop(ctrl.npcID, goods))
+	if ctrl.stateTracker.performInterrupt() {
+		ctrl.goods = goods
+		ctrl.conn.Send(packetNpcShop(ctrl.npcID, goods))
+		ctrl.vm.Interrupt("SendShop")
```
Hucaru, line 680: I assume the client did send a termination packet?

ErwinsExpertise: it did

channel/script.go
```diff
@@ -508,48 +674,206 @@ func (ctrl *npcChatController) SendGuildEmblemEditor() {
...
 func (ctrl *npcChatController) Selection() int32 {
-	val := ctrl.stateTracker.selections[ctrl.stateTracker.selection]
-	ctrl.stateTracker.selection++
-	return val
+	if len(ctrl.stateTracker.selections) == 0 {
+		return -1
+	}
+
+	if len(ctrl.stateTracker.selections) == 1 {
+		ctrl.selectionCalls++
```
Hucaru: Can you walk through an example of why this variable is needed? Unless I am missing something you are always returning the last item in the selections slice? Previously it was:
```
func (ctrl *npcChatController) Selection() int32 {
    val := ctrl.stateTracker.selections[ctrl.stateTracker.selection]
    ctrl.stateTracker.selection++ // tracks which selection result we want next
    return val
}
```
which would return the relevant selection for where you were in the script

Looking at this line https://github.com/ErwinsExpertise/Valhalla/blob/master/scripts/npc/1012103.js#L29 I am not following where you mention that the user needs to make 2 selection in the UI as it appears `npc.selection` is called once per avatar ui window.

channel/script.go
```diff
@@ -508,48 +674,206 @@ func (ctrl *npcChatController) SendGuildEmblemEditor() {
...
 
 // InputString value
 func (ctrl *npcChatController) InputString() string {
-	val := ctrl.stateTracker.inputs[ctrl.stateTracker.input]
-	ctrl.stateTracker.input++
-	return val
+	if len(ctrl.stateTracker.inputs) == 0 {
+		return ""
+	}
+	return ctrl.stateTracker.inputs[len(ctrl.stateTracker.inputs)-1]
```
Hucaru: You don't want the last item inserted as if you have a nested set of inputs then when the script executes again (from the start) after the interrupt it will take the answer to the last input as the answer for all. This applies to all the user inputs, which is why I had a counter variable for all of them to know which result to return when the script asked for it:

```

func (ctrl *npcChatController) Selection() int32 {
	val := ctrl.stateTracker.selections[ctrl.stateTracker.selection]
	ctrl.stateTracker.selection++
	return val
}

func (ctrl *npcChatController) InputString() string {
	val := ctrl.stateTracker.inputs[ctrl.stateTracker.input]
	ctrl.stateTracker.input++
	return val
}

func (ctrl *npcChatController) InputNumber() int32 {
	val := ctrl.stateTracker.numbers[ctrl.stateTracker.number]
	ctrl.stateTracker.number++
	return val
}
```

channel/script.go
ErwinsExpertise: It was more or less just leftover from experimenting, and extra selectionCalls variable isn’t needed. I’ve reverted Selection() to use the existing per-input counter on stateTracker (selection) so we always return the next selection in order across interrupts. The current method is needed to support both single-step interactions and multi-step flows without heuristics or NPC-specific logic.

channel/script.go
ErwinsExpertise: Returning the last value is wrong for nested/multi-step prompts because after an interrupt the script restarts and would keep reading the same value. I’ve switched these to use the per-input counters (input/number) just like Selection(). This makes the engine consume responses in sequence and keeps multi-step scripts consistent.

Hucaru (approved):

Issue #90, opened 2025-09-03
Add Some Commands for Testing Purposes


LuneArk:
For testing purposes, here are some commands that I believe would be useful:

*The commands are stylized in camel case for readability, but making them case-insensitive might be better.

- **/setMaxHP ["amount"]:** Sets the character's maximum HP  to a desired amount. The existing command `/hp` is primarily intended for setting your current HP and does not allow you to lower your maximum HP below what you have previously set with that command, making it irreversible. For instance, if you first use **"/hp 500"** and then **"/hp 100"**, your HP will read 100/500.

- **/setMaxMP ["amount"]:** Sets the character's maximum MP  to a desired amount. The existing command `/mp` is primarily intended for setting your current MP and does not allow you to lower your maximum MP below what you have previously set with that command, making it irreversible. For instance, if you first use **"/mp 500"** and then **"/mp 100"**, your MP will read 100/500.

- **/str, /dex, /int, /luk:** Sets your desired stat amount.

- **/skillLv ["Skill ID / Skill Name", "Skill Level" / "max"]:** Adjusts the level of a specific skill to the desired amount or maxes it.

- **/maxSkills:** Maxes all skills across all Classes and their Job Advancements.

- **/resetSkills:** Removes all assigned skill points across all Classes and their Job Advancements.

- **/questFinish ["Quest ID"]:** Completes an entire quest line.

- **/questUntil ["Quest ID", "Quest Part #"]:** Completes up to a specific part within a quest line.

- **/questReset:** Resets a quest line as if it has never been completed.

- **/removeTimer:** Removes timers on maps, such as in Party Quests, Boss Maps, or Monster Maps.

- **/clearDrops:** Remove all items that are on the ground.

---

I'm not sure if this project intends to use some of the older OdinMS command names, but I believe simplifying them to be more intuitive, shorter, and easier to use would enhance testing efficiency. Here are some optional suggested changes:

- Change `/killmobs` -> `/killAll`: Kills all monsters on the map.

- Change `/spawnmob` -> `/spawn`: Spawns a specific monster in the given quantity based on the Monster ID.

PR #113, opened 2025-09-17
Feature/Add gm commands

ErwinsExpertise:
This PR will address #90 and add the requested GM commands.

### DIFF
--- a/channel/commands.go
+++ b/channel/commands.go
@@ -7,7 +7,9 @@ import (
 	"math"
 	"strconv"
 	"strings"
+	"time"
 
+	"github.com/Hucaru/Valhalla/constant"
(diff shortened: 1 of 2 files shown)


Hucaru (approved):

channel/commands.go
```diff
@@ -314,6 +316,342 @@ func (server *Server) gmCommand(conn mnet.Client, msg string) {
...
+		}
+
+		if val < 0 {
+			conn.Send(packetMessageRedText("Stat cannot be negative"))
+			return
+		}
+
+		switch command[0] {
+		case "str":
+			target.str = int16(val)
```
Hucaru, line 406: Minor: Should these be refactored as player struct member variables? Do we forsee needing to update this value from multiple locations?

PR #114, opened 2025-09-18
Add Player Storage

ErwinsExpertise:
This PR will implement player storage #103

couple npc fixes in there too

### DIFF
--- a/scripts/npc/2012013.js
+++ b/scripts/npc/2012013.js
@@ -3,9 +3,9 @@ var TICKET_ID = 4031074;
 var BOARD_MAP = 200000122;
 
 var props = plr.instanceProperties();
-var salesOpen = ("canSellTickets" in props) && props["canSellTickets"];
+var boardingOpen = ("canBoard" in props) && props["canBoard"];
 
-if (!salesOpen) {
+if (!boardingOpen) {
     npc.sendOk("Boarding is not available right now. Please wait for the next boarding announcement.");
 } else if (plr.itemCount(TICKET_ID) < 1) {
     npc.sendOk(
--- a/scripts/event/boats.js
+++ b/scripts/event/boats.js
@@ -45,11 +45,11 @@ function showBoats(controller, maps, show, type) {
     }
 }
 
-function allowTicketSales(controller, maps, allow) {
+function allowBoarding(controller, maps, allow) {
     for (var i = 0; i < maps.length; i++) {
         var field = controller.field(maps[i])
         for (var j = 0; j < field.instanceCount(); j++) {
-            field.getProperties(j)["canSellTickets"] = allow
+            field.getProperties(j)["canBoard"] = allow
         }
     }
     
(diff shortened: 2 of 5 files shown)


channel/item.go
```diff
@@ -468,6 +468,63 @@ func (v Item) bytes(shortSlot bool) []byte {
 	return p
 }
 
+func (v Item) storageBytes() []byte {
```
Hucaru, line 471: This function looks very similar to https://github.com/Hucaru/Valhalla/blob/master/channel/item.go#L393 I think the item `bytes()` should be modified to support both methods of encoding as it's preferable to not create multiple "serialize" functions for types if we can avoid it.

channel/npc.go
```diff
@@ -246,24 +246,87 @@ func packetTradeError() mpacket.Packet {
 	return packetNpcShopResult(0xFF)
 }
 
+func storageFlagForInv(inv byte) uint16 {
+	switch inv {
+	case 1:
```
Hucaru, line 251: Add constants for these magic numbers to align with the handler code in this PR and going forward we should do this as much as possible.

channel/storage.go
```diff
@@ -0,0 +1,296 @@
...
+)
+
+// Storage capacity bounds
+const (
+	storageMinSlots byte = 20
+	storageMaxSlots byte = 255
+)
+
+type storage struct {
+	accountID int32
```
Hucaru, line 18: I would remove the account ID from here since this struct is a member of the player struct, you can then pass the accountID to the save/load functions and save yourself the accountID value check.

scripts/npc/1043000.js
Hucaru, line 1: What are these NPCs from? I don't remember encountering these back in the day.

Hucaru:
Don't forget to update README.

scripts/npc/1043000.js
ErwinsExpertise: They're from quests, like from sabitrama's herbal medicine.

scripts/npc/2012013.js
```diff
@@ -1,58 +1,9 @@
...
-
-    } else {                  // ride the ship
-        if (npc.sendYesNo("Would you like to board the ship to Ludibrium now?")) {
-            plr.warp(200090100, 0);
-            plr.startMapTimeLimitTask(60, 220000110);
-        } else {
-            npc.sendOk("Do you have some business you need to take care of here?");
-        }
-    }
+if ("canSellTickets" in props && props["canSellTickets"]) {
```
Hucaru: Minor: I think ticket sales happen before boarding is allowed? If so the boat script should probably set a new instance property for this.

ErwinsExpertise: yeah I believe ticket sales are any time. The timed thing should be boarding. Maybe we switch it it `canBoard` ?

ErwinsExpertise: Okay @Hucaru I refactored to be `canBoard` and `allowBoarding` instead:
https://github.com/Hucaru/Valhalla/pull/114/commits/2fe5fd66ad78f6c6b9e133daa4a0ffed5d400b8f

That should better align with what we expect. Let me know if it's good to merge

Hucaru (approved):

PR #115, opened 2025-09-20
Feature: Add Messenger

ErwinsExpertise:
This PR will implement the in-game messenger resolving #112

messenger

### DIFF
--- a/internal/serialisable_data.go
+++ b/internal/serialisable_data.go
@@ -128,3 +128,8 @@ type KV struct {
 	K byte
 	V int32
 }
+
+func (kv *KV) Serialise(p *mpacket.Packet) {
+	p.WriteByte(kv.K)
+	p.WriteInt32(kv.V)
+}
--- a/channel/message.go
+++ b/channel/message.go
@@ -339,13 +339,11 @@ func packetMessengerEnter(slot, gender, skin, ch byte, face, hair, cashW, petAcc
 	p.WriteBool(true)
 	p.WriteInt32(hair)
 	for _, kv := range vis {
-		p.WriteByte(kv.K)
-		p.WriteInt32(kv.V)
+		kv.Serialise(&p)
 	}
 	p.WriteInt8(-1)
 	for _, kv := range hid {
-		p.WriteByte(kv.K)
-		p.WriteInt32(kv.V)
+		kv.Serialise(&p)
 	}
 	p.WriteInt8(-1)
 	p.WriteInt32(cashW)
@@ -366,13 +364,11 @@ func packetMessengerAvatar(slot, gender, skin byte, face, hair, cashW, petAcc in
 	p.WriteBool(true)
 	p.WriteInt32(hair)
 	for _, kv := range vis {
-		p.WriteByte(kv.K)
-		p.WriteInt32(kv.V)
+		kv.Serialise(&p)
 	}
 	p.WriteInt8(-1)
 	for _, kv := range hid {
-		p.WriteByte(kv.K)
-		p.WriteInt32(kv.V)
+		kv.Serialise(&p)
 	}
 	p.WriteInt8(-1)
 	p.WriteInt32(cashW)
--- a/internal/packets.go
+++ b/internal/packets.go
@@ -329,13 +329,11 @@ func PacketWorldMessengerEnter(recipientID, face, hair, cashW, petAcc int32, slo
 	p.WriteBool(true)
 	p.WriteInt32(hair)
 	for _, kv := range vis {
-		p.WriteByte(kv.K)
-		p.WriteInt32(kv.V)
+		kv.Serialise(&p)
 	}
 	p.WriteInt8(-1)
 	for _, kv := range hid {
-		p.WriteByte(kv.K)
-		p.WriteInt32(kv.V)
+		kv.Serialise(&p)
 	}
 	p.WriteInt8(-1)
 	p.WriteInt32(cashW)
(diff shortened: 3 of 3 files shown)


channel/handlers.go
```diff
@@ -4336,3 +4533,210 @@ func (server *Server) playerUseStorage(conn mnet.Client, reader mpacket.Reader)
...
+
+func (server *Server) playerHandleMessenger(conn mnet.Client, reader mpacket.Reader) {
+	plr, err := server.players.getFromConn(conn)
+	if err != nil {
+		return
+	}
+
+	mode := reader.ReadByte()
+
+	p := mpacket.CreateInternal(opcode.ChannelPlayerMessengerEvent)
```
Hucaru: Internal packets should exist in: https://github.com/Hucaru/Valhalla/blob/master/internal/packets.go

channel/handlers.go
```diff
@@ -3632,6 +3637,198 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
...
+			plr.Send(packetMessengerSelfEnter(slot))
+		}
+	case constant.MessengerEnterResult:
+		recipientID := reader.ReadInt32()
+		slot := reader.ReadByte()
+
+		gender := reader.ReadByte()
+		skin := reader.ReadByte()
+		face := reader.ReadInt32()
+		_ = reader.ReadBool()
```
Hucaru, line 3656: Do we know what this bool represents?

channel/handlers.go
```diff
@@ -3632,6 +3637,198 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
...
+		hair := reader.ReadInt32()
+
+		vis := make([]struct {
+			k byte
+			v int32
+		}, 0, 16)
+		for {
+			b := reader.ReadByte()
+			if int8(b) == -1 {
```
Hucaru: multiple: `if b == 0xFF` ?

channel/handlers.go
```diff
@@ -3632,6 +3637,198 @@ func (server Server) handleChatEvent(conn mnet.Server, reader mpacket.Reader) {
...
+		}
+		cashW := reader.ReadInt32()
+		petAcc := reader.ReadInt32()
+
+		name := reader.ReadString(reader.ReadInt16())
+		ch := reader.ReadByte()
+		announce := reader.ReadBool()
+
+		if plr, err := server.players.getFromID(recipientID); err == nil {
+			p := mpacket.CreateWithOpcode(opcode.SendChannelMessenger)
```
Hucaru: Packet definition shouldn't be embedded in the handler

channel/handlers.go
```diff
@@ -4336,3 +4533,210 @@ func (server *Server) playerUseStorage(conn mnet.Client, reader mpacket.Reader)
...
+		p.WriteInt8(-1)
+		p.WriteInt32(cashWeapon)
+		p.WriteInt32(petAccessory)
+
+	case constant.MessengerLeave:
+	case constant.MessengerInvite:
+		invitee := reader.ReadString(reader.ReadInt16())
+		p.WriteString(invitee)
+	case constant.MessengerBlocked:
+		invitee := reader.ReadString(reader.ReadInt16())
```
Hucaru, line 4552: I would prefer it if the reading and writing of packets was seperated out and not interweaved

world/handlers.go
```diff
@@ -475,3 +477,372 @@ func (server *Server) handleGuildEvent(conn mnet.Server, reader mpacket.Reader)
...
+func (m *messengerRoom) findMemberSlot(id int32) (byte, bool) {
+	for i, mm := range m.members {
+		if mm != nil && mm.charID == id {
+			return byte(i), true
+		}
+	}
+	return 0, false
+}
+
+var messengerRooms = map[int32]*messengerRoom{}
```
Hucaru: Instead of a global it would be consistent to have it be a member variable of the server struct

world/handlers.go
```diff
@@ -475,3 +477,372 @@ func (server *Server) handleGuildEvent(conn mnet.Server, reader mpacket.Reader)
...
+		}
+
+	default:
+		log.Println("Unknown messenger op (channel->world):", mode)
+	}
+}
+
+// ===== World->Channel packets for Messenger (inline AvatarLook fields) =====
+
+func packetWorldMessengerSelfEnter(recipientID int32, slot byte) mpacket.Packet {
```
Hucaru: These packets should live with the other internal packets

internal/serialisable_data.go
```diff
@@ -123,3 +123,8 @@ func (party *Party) SerialisePacket(reader *mpacket.Reader) {
 		party.Level[i] = reader.ReadInt32()
 	}
 }
+
+type KV struct {
+	K byte
+	V int32
+}
```
Hucaru, line 130: It looks like every instance of KV can be serialised as `[u8, u32]` so I think considering it's in the `serialisable_data` file to give it a Serialise where it takes a packet as an arg e.g.

```
func (kv KV) Serialise(p *mpacket.Packet) {
    p.WriteByte(kv.K)
    p.WriteInt32(kv.V)
}
```
and replace all of the following with the above:
```
    for _, kv := range vis {
		p.WriteByte(kv.K)
		p.WriteInt32(kv.V)
	}
```

Just realised that the functions in this file named `SerialisePacket` should be named `DeserialisePacket` instead🤦

ErwinsExpertise: @Hucaru good idea, made those changes

PR #119, opened 2025-09-24
Reactors Drops and Warp

ErwinsExpertise:
This PR refactors reactors to perform drops, and warping. It also makes the `processStateSideEffects` a little more intuitive.

### DIFF
--- a/channel/pools.go
+++ b/channel/pools.go
@@ -1094,7 +1094,7 @@ type reactorTableEntry map[string]interface{}
 
 var reactorTable map[string]map[string]reactorTableEntry
 
-func PopulateReactorTable(reactorJSON string) error {
+func populateReactorTable(reactorJSON string) error {
 	f, err := os.Open(reactorJSON)
 	if err != nil {
 		return err
@@ -1115,7 +1115,7 @@ type reactorDrops struct {
 
 var reactorDropTable map[string]reactorDrops
 
-func PopulateReactorDropTable(reactorJSON string) error {
+func populateReactorDropTable(reactorJSON string) error {
 	b, err := os.ReadFile(reactorJSON)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


server_channel.go
```diff
@@ -91,6 +91,13 @@ func (cs *channelServer) run() {
 	elapsed = time.Since(start)
 	log.Println("Loaded and parsed reactor data in", elapsed)
 
+	start = time.Now()
```
Hucaru: This could be moved into the `Initialise` method of the server struct in the channel package and you wouldn't then need to export the populate function. The path to the file should probably be passed in case it's made configurable in the future.

channel/pools.go
```diff
@@ -1108,26 +1108,51 @@ func PopulateReactorTable(reactorJSON string) error {
...
+type reactorDrops struct {
+	items []int
+	money int
 }
 
-func (r *fieldReactor) calcEventRect() (rect, bool) {
-	st, ok := r.info.States[int(r.state)]
-	if !ok || len(st.Events) == 0 {
-		return rect{}, false
+var reactorDropTable map[string]reactorDrops
```
Hucaru, line 1116: To keep it in line with the rest of the package this should be a member varaible of the server struct

ErwinsExpertise: This feature is modeled more closely after drops and reactors, which are not managed as member variables on the server struct. Reactors were actually modeled off of drops, so for consistency it made sense to follow that same pattern here.

If the consensus is that this kind of state should live on server, we should probably refactor drops and reactors as well so everything follows the same convention. Otherwise, I think it makes sense to keep this consistent with the existing drop/reactor handling.

Hucaru: In that case leave it as is. In a future change they can all be moved over to the server struct for consistency.

ErwinsExpertise: Sounds good. Rerequested approval

Hucaru: I have a previous comment regarding where the populate function should be called from.

server_channel.go
ErwinsExpertise: @Hucaru I made these changes, and I also did the same for drops, and reactors to ensure they're all following the same pattern.

Hucaru (approved):

Issue #120, opened 2025-09-28
Update Dockerfile for compose


ErwinsExpertise:
The Dockerfile under /docker is out of date. It needs to be updated to match the one in root

trameathia:
after moving the root Dockerfile under /docker it then throws an error trying to copy Data.nx
```
Step 10/17 : COPY Data.nx /app/Data.nx
COPY failed: file not found in build context or excluded by .dockerignore: stat Data.nx: file does not exist
ERROR: Service 'login_server' failed to build : Build failed
```

Issue #118, opened 2025-09-24
docker-compose.yaml really should be docker-compose.yml


halogod35:
Most things docker related tend to default to .yml. Lots of user errors and bug reports can be avoided by renaming .yaml to .yml

<img width="497" height="104" alt="Image" src="https://github.com/user-attachments/assets/565f13c7-6067-4489-8e2b-bcabd7f5d540" />

Even portainer defaults to .yml

PR #121, opened 2025-10-01
Add pets

### LINKED ISSUE #102: Implement Pets

ErwinsExpertise:
This PR will add pets. 

Resolves #102 #120  #118

Screenshot_2

pickup

### DIFF
--- a/channel/field.go
+++ b/channel/field.go
@@ -827,7 +827,7 @@ func packetMapPlayerEnter(plr *Player) mpacket.Packet {
 	} else {
 		p.WriteBool(false)
 	}
-	
+
 	if plr.petCashID != 0 {
 		plr.pet.pos = plr.pos
 		p.WriteBool(true)
--- a/channel/handlers.go
+++ b/channel/handlers.go
@@ -474,13 +474,23 @@ func (server Server) playerRequestAvatarInfoWindow(conn mnet.Client, reader mpac
 }
 
 func (server Server) playerPassiveRegen(conn mnet.Client, reader mpacket.Reader) {
-	reader.ReadBytes(4) //?
+	flag := reader.ReadInt32()
 
-	hp := reader.ReadInt16()
-	mp := reader.ReadInt16()
+	var hp, mp int16
+
+	if flag&0x0400 != 0 {
+		hp = reader.ReadInt16()
+	} else {
+		hp = 0
+	}
+
+	if flag&0x1000 != 0 {
+		mp = reader.ReadInt16()
+	} else {
+		mp = 0
+	}
 
 	player, err := server.players.getFromConn(conn)
-
 	if err != nil {
 		return
 	}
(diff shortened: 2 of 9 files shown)


channel/pets.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	successProb := float64(react.Prob) * ((elapsed/10_000.0)*0.01 + 1) * mult
+	success := float64(rand.Intn(100)) < successProb
+	if success {
+		increasePetCloseness(plr, pet, int16(react.Inc))
+	}
+	return success
+}
+
+func increasePetCloseness(plr *Player, pet *pet, inc int16) {
```
Hucaru: Does this need to be it's own function? It's very small and as far as i can tell it's only used by the Interact function.

channel/pets.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func (p *pet) updateMovement(frag movementFrag) {
+	p.pos.x = frag.x
+	p.pos.y = frag.y
+	p.pos.foothold = frag.foothold
+	p.stance = frag.stance
+}
+
+func handlePetInteraction(plr *Player, pet *pet, interactionID byte, multiplier bool) bool {
```
Hucaru, line 93: Can a playerr interact with a pet that is not theirs?

channel/handlers.go
```diff
@@ -4625,3 +4643,139 @@ func (server *Server) playerHandleMessenger(conn mnet.Client, reader mpacket.Rea
...
+	plr.Send(packetPetInteraction(plr.ID, interactionID, success, false))
+}
+
+func (server *Server) playerPetLoot(conn mnet.Client, reader mpacket.Reader) {
+	plr, err := server.players.getFromConn(conn)
+	if err != nil || plr.pet == nil || !plr.pet.spawned {
+		return
+	}
+
+	reader.Skip(4) // unused pos
```
Hucaru, line 4769: Which pos is this? It could be useful in preventing pets from vacuuming all items on a field.

channel/field.go
```diff
@@ -817,6 +821,20 @@ func packetMapPlayerEnter(plr *Player) mpacket.Packet {
 	p.WriteInt16(plr.pos.y)
 	p.WriteByte(plr.stance)
 	p.WriteInt16(plr.pos.foothold)
+
+	p.WriteBool(false) // GM Hide
```
Hucaru: Should this not check if plr has gm hide buf on?

channel/pets.go
ErwinsExpertise: They cannot. The handler is only triggered by a client speaking to their pet. As an example saying "bad"

channel/handlers.go
ErwinsExpertise: It's the player's pos, which is kinda useless for this. Updated to check pet POS against drop POS.

channel/field.go
ErwinsExpertise: added check

channel/pets.go
Hucaru: Is there a need to pass in a pet ptr then if it's always the plr's pet?

Hucaru (approved):

PR #124, opened 2025-10-05
added support for ENV variables

trameathia:
switched from github.com/BurntSushi/toml to github.com/spf13/viper since this project supports TOML and ENV vars for configurations.  Added mapping for ENV variables and updated the docker-compose.yml to have all the ENVs in an easy to edit place


server_config.go
```diff
@@ -1,110 +1,155 @@
...
 
-	if _, err := toml.DecodeFile(fname, config); err != nil {
-		log.Fatal(err)
+	bindEnvs(v, reflect.TypeOf(fullConfig{}), nil, "VALHALLA")
+
+	var config fullConfig
+	if err := v.Unmarshal(&config); err != nil {
+		log.Fatalf("failed to unmarshal config: %v", err)
 	}
+	log.Printf("World → login %q:%q | listen %q:%q", config.World.LoginAddress, config.World.LoginPort, config.World.ListenAddress, config.World.ListenPort)
```
Hucaru: Won't this log for all server types? Not sure if there is much meaning in this message for the channel/cash shop server. 

If you still want this information logged then I think it should be after the following line for each server type with their relevant information.

https://github.com/Hucaru/Valhalla/blob/master/server_login.go#L64

trameathia: Oops thought I removed all the extra log statements I had added to find problems, looks like I missed one.  I'll push up a change after work.

trameathia: i removed the bad log statement and added new ones where you recommended to each server type so we could find that info in the logs.

trameathia:
@Hucaru It's ready for review again.

Hucaru (approved):

PR #127, opened 2025-10-07
Add trade and bug fixes

ErwinsExpertise:
This PR will add player trade and also fixes a couple bugs. 

Screenshot_3

### DIFF
--- a/channel/player.go
+++ b/channel/player.go
@@ -614,7 +614,7 @@ func (d *Player) SetMaplePoints(points int32) {
 	d.MarkDirty(DirtyMaplePoints, 300*time.Millisecond)
 }
 
-// GiveItem grants the given item to a player and returns the db ID
+// GiveItem grants the given item to a player and returns the item
 func (d *Player) GiveItem(newItem Item) (error, Item) { // TODO: Refactor
 	isRechargeable := func(itemID int32) bool {
 		base := itemID / 10000


constant/constants.go
```diff
@@ -12,6 +12,7 @@ const (
 
 const (
 	MaxItemStack = 200
+	MaxMesos     = 2147483647
```
Hucaru: `math.MaxInt32` avoids any potential typos

internal/packets.go
```diff
@@ -489,3 +489,8 @@ func PacketMessengerAvatar(gender, skin, channelID byte, charID, face, hair, cas
 	p.WriteInt32(petAccessory)
 	return p
 }
+
+func PacketRequestChannelInfo() mpacket.Packet {
```
Hucaru: Rename to `PacketCashShopRequestChannelInfo` as it's not clear from the name which server this is for.

channel/handlers.go
```diff
@@ -2757,20 +2764,75 @@ func (server Server) roomWindow(conn mnet.Client, reader mpacket.Reader) {
...
 	case roomMesos:
-		// amount := reader.ReadInt32()
+		amount := reader.ReadInt32()
+
+		if plr.mesos < amount {
+			plr.Send(packetPlayerNoChange())
+			return
+		}
+
+		plr.takeMesos(amount)
```
Hucaru, line 2814: We shouldn't be taking any items/mesos at this stage. The players should only loose anything when the trade completes, this will simplify trade failures as there is nothing to return and prevents issues when a player disconnects mid trade causing potential item loss.

The trade should keep track of the inv type, slot, amount and mesos to trade and when agreed perform the item/mesos take/give. The rollback method can be removed as there is nothing to rollback and failures are just trade packet messages as the client inventory state has not been changed.

constant/constants.go
```diff
@@ -157,6 +158,130 @@ const (
 	PetRemoveExpire byte = 2
 )
 
+const (
+	MiniRoomCreate        byte = 0
```
Hucaru, line 161: Thanks for pulling these constants out but I can't see the room.go changes associated with a lot of these?

channel/handlers.go
ErwinsExpertise: I understand the reasoning behind deferring item and meso removal until trade completion, as it can simplify rollback handling. However, I think this approach introduces some practical issues that may make the system less reliable overall.

If we don’t remove the items or mesos right away, they would still exist in the player’s inventory until the trade is finalized. That creates opportunities for inconsistencies if a player performs other actions during the trade or disconnects right after accepting, potentially throwing off backend sync or database writes. It could also open up edge cases where items appear duplicated or out of sync with the client state.

The current approach, where the server takes the items immediately, keeps the state authoritative and consistent throughout the trade. It also follows the same pattern used in other inventory operations, which helps avoid desyncs if a player cancels or disconnects mid-trade. I think this provides stronger consistency guarantees, prevents potential duping or desync scenarios, and keeps the trade logic predictable and robust.

constant/constants.go
ErwinsExpertise: Didn't actually make those changes, but just pushed a commit with them.

channel/handlers.go
Hucaru: In that case can we remove the items/mesos from the player but not send the packets to the client? This would keep the server authoritative during the whole trade process but not send any more packets than needed to the client (I am assuming the client figures out which item to visually remove from the inventory when it receives a trade window insert packet). Furthermore I think any acts of moving items/mesos should be done after the room has been found not before as it might compound bugs.

I wonder if instead of storing the mesos and items in a brand new nested map in the room it would probably be simpler and faster to have a trade escrow struct in the plr struct that stores the items and mesos that are part of the transaction. The items can be stored in a slice with a capacity that is the same size as the amount of items allowed for trade hopefully being stack allocated instead of heap (who knows what the runtime will actually do). This will alsow make handling a client disconnect easier as the items and mesos in escrow can be handled as part of the save.

ErwinsExpertise: The client doesn’t actually handle inventory updates automatically when receiving the trade insert packet — it just displays whatever the server encodes. If we don’t send the packet to remove the item, it’ll still appear in the player’s inventory, which leads to desync between client and server state. That’s why we currently take and send it right away: it keeps the two views consistent and avoids duping or mismatched item states.

We could technically delay committing the changes to the database until the trade is accepted, but that would require extra handling around pending item state and add more branching logic for both success and rollback cases. At that point, it becomes more complex than the current approach and still doesn’t provide much benefit in terms of simplicity or reliability.

Hucaru: Ah if we have to communicate to the client the removal of the item/mesos from the inventory window then so be it. 

If the trade is taking a long time and the player is updated in the DB after moving the item into the trade window and then they disconnect before completing the trade does the rollback method get called? Currently I don't think it does which would lead to permanent item/mesos loss. Attaching the escrow to the specific players rather than the rooms would allow this to be handled by the plr logout method and also save on having to use nested maps in the room struct. This means the only state the room struct needs to hold is if both sides agree to the transaction and then move the items/mesos in each player's escrow to the other player's inventory.

ErwinsExpertise: That’s a good point about disconnects. I think we can handle that cleanly without moving the escrow into the player, since all of the trade state already lives in memory within the room. We can hook into the disconnect flow and have the room trigger a rollback before it’s cleaned up, which keeps the trade fully self-contained and avoids having to coordinate teardown across multiple player structs. The current tradeRoom layout already fits this model well — it’s minimal, isolated, and easy to maintain.

I just pushed a commit that adds proper disconnect handling and rollback logic. It also includes stricter validation to ensure players have enough space for items and mesos, and the trade finalization now runs atomically to prevent any inconsistencies or edge cases.

channel/server.go
```diff
@@ -319,6 +321,43 @@ func (server *Server) ClientDisconnected(conn mnet.Client) {
...
+		return
+	}
+
+	if trade, ok := r.(*tradeRoom); ok {
+		trade.removePlayer(plr)
+		_ = pool.removeRoom(trade.id())
+		return
+	}
+
+	if game, ok := r.(gameRoomer); ok {
```
Hucaru: I think we can move this function logic into https://github.com/Hucaru/Valhalla/blob/master/channel/pools.go#L741 as your game room logic is almost identical to what I wrote.

The removePlayer function for the field calls the instance removePlayer which in turn calls the removePlayer function for all the pools incl. the room pool.

channel/room.go
```diff
@@ -743,59 +709,171 @@ func (r *memoryRoom) shuffleCards() {
...
 }
 
 func (r *tradeRoom) removePlayer(plr *Player) {
-	// Note: since anyone leaving the room causes it to close we don't need to remove players
-	for i, v := range r.players {
-		if v.Conn != plr.Conn {
-			v.Send(packetRoomLeave(byte(i), 0x02))
-		}
-	}
+	r.closeWithReason(constant.RoomLeaveTradeCancelled, true)
```
Hucaru, line 742: We should change the if statement to a type switch: https://github.com/Hucaru/Valhalla/blob/master/channel/pools.go#L741 and call this removePlayer function

ErwinsExpertise: I like the idea of using a switch instead. Pushed a commit

channel/server.go
```diff
@@ -336,26 +336,7 @@ func (server *Server) cleanupMiniRoom(plr *Player) {
...
-
-	if game, ok := r.(gameRoomer); ok {
-		game.kickPlayer(plr, 0x00)
-		if r.closed() {
-			_ = pool.removeRoom(r.id())
-		} else {
-			pool.updateGameBox(r)
-		}
-	}
+	inst.roomPool.removePlayer(plr)
```
Hucaru: This should not be needed as instances manage their pools directly and already removePlayers when they leave the instance: https://github.com/Hucaru/Valhalla/blob/master/channel/field.go#L617 which gets called when a client disconnects from https://github.com/Hucaru/Valhalla/blob/master/channel/server.go#L283

ErwinsExpertise: Ah yeah, I forgot that was being called there when I added tradeRoom type into remove player. Removed and pushed new commit.

channel/room.go
```diff
@@ -743,59 +709,171 @@ func (r *memoryRoom) shuffleCards() {
...
+	if !p1.canReceiveItems(changes[0].itemsToGive) {
+		r.closeWithReason(constant.MiniRoomTradeInventoryFull, true)
+		return
+	}
+	if !p2.canReceiveItems(changes[1].itemsToGive) {
+		r.closeWithReason(constant.MiniRoomTradeInventoryFull, true)
+		return
+	}
+
+	if changes[0].mesosChange > 0 {
```
Hucaru: Should the mesos change happen after item change considering that could fail whilst the mesos passes?

ErwinsExpertise: We are checking whether or not **all** operations can occur prior to performing any of the exchanges, so it really doesn't matter which happens first. If any of the 4 checks fail then we cancel the trade and rollback.

channel/room.go
```diff
@@ -743,59 +709,171 @@ func (r *memoryRoom) shuffleCards() {
 
 type tradeRoom struct {
 	room
+	mesos     map[int32]int32
+	items     map[int32]map[byte]Item
```
Hucaru, line 713: Does this need to be nested map? Could it be `map[int32][max slot count]Item`?

ErwinsExpertise: The nested map is intentional. The inner map is keyed by the trade slot (byte), which isn’t guaranteed to be sequential or contiguous. Using a fixed array or slice would require us to manage placement manually and handle unused indexes.

For example:
```
func (r *tradeRoom) insertItem(tradeSlot byte, plrID int32, item Item) {
	r.items[plrID][tradeSlot] = item
}
```

Here, we directly index by the slot value sent from the client. With a map, this is a single, clean operation. If we used an array or slice, we’d have to ensure the slot fits within bounds and handle empty slots explicitly, which adds unnecessary complexity for no performance gain. The map keeps the structure flexible, readable, and easy to maintain.

Hucaru: Is there a check for tradeSlot being a valid slot e.g. 99?

ErwinsExpertise: Pushed a check for valid slot

channel/room.go
```diff
@@ -749,6 +751,16 @@ func (r tradeRoom) reject(code byte, name string) {
 }
 
 func (r *tradeRoom) insertItem(tradeSlot byte, plrID int32, item Item) {
```
Hucaru, line 753: Should this function return a boolean and let the calling handler inform the client that an error occurred?

channel/handlers.go
```diff
@@ -2763,15 +2770,69 @@ func (server Server) roomWindow(conn mnet.Client, reader mpacket.Reader) {
...
+			plr.Send(packetPlayerNoChange())
+			return
+		}
+
+		newItem := item
+		newItem.amount = amount
+
+		if r, err := pool.getPlayerRoom(plr.ID); err == nil {
+			if tr, ok := r.(*tradeRoom); ok {
+				tr.insertItem(tradeSlot, plr.ID, newItem)
```
Hucaru, line 2800: If this function is made to return a boolean as suggested in the other comment then we can send the client red text informing them an error occured. If we want it to be descriptive we could return an error type instead of boolean and forward that to the client as red text

ErwinsExpertise: **Technically**, the actions that would cause this function to return are not actions that a player should be able to do in the first place. If these instances are occurring then the user is hacking, and I'm not sure that's something we'd want to relay to them. Since there aren't any anti-cheat/ban mechanisms in the source, this is the only real option is to log it.

Hucaru: Fair enough

Issue #134, opened 2025-10-11
enter the cash shop messes up the display in your quest log for kill x monster quests


trameathia:
when doing the quest to kill green snail, blue snails, and shrooms i had multiple of each type killed.  when i went into the cash shop and came back out my quest log said 0/10 for all of them.  after i killed 1 enemy they then displayed again but i don't think that 1 kill counted towards my totals.

PR #137, opened 2025-10-12
bug fixes

### LINKED ISSUE #135: meso magent also working as item pouch
i had a husky with a meso magnet and they were picking up items as well even though i didn't buy the item pouch

ErwinsExpertise:
This PR will resolve #135 and #134

### DIFF
--- a/channel/handlers.go
+++ b/channel/handlers.go
@@ -534,7 +534,7 @@ func (server Server) playerUseChair(conn mnet.Client, reader mpacket.Reader) {
 	chairID := reader.ReadInt32()
 	plr.chairID = chairID
 
-	plr.inst.send(packetPlayerShowChair(plr.ID, chairID))
+	plr.inst.sendExcept(packetPlayerShowChair(plr.ID, chairID), plr.Conn)
 	plr.Send(packetPlayerChairUpdate())
 }
 
@@ -548,7 +548,7 @@ func (server Server) playerStand(conn mnet.Client, reader mpacket.Reader) {
 	chairIndex := reader.ReadInt16()
 
 	plr.chairID = 0
-	plr.inst.send(packetPlayerShowChair(plr.ID, 0))
+	plr.inst.sendExcept(packetPlayerShowChair(plr.ID, 0), plr.Conn)
 	plr.Send(packetPlayerChairResult(chairIndex))
 	plr.Send(packetPlayerChairUpdate())
 }


channel/player.go
```diff
@@ -2065,6 +2065,37 @@ func (p *Player) updatePet() {
 	p.inst.send(packetPlayerPetUpdate(p.pet.sn))
 }
 
+func (p *Player) petCanTakeDrop(drop fieldDrop) bool {
+	if p.pet == nil {
+		return false
+	}
+
+	if drop.mesos > 0 {
+		if p.hasEquipped(1812000) {
```
Hucaru: These should be extracted as mesos magnet and item pouch constants

channel/handlers.go
```diff
@@ -507,16 +525,54 @@ func (server Server) playerPassiveRegen(conn mnet.Client, reader mpacket.Reader)
...
+
+	if healAmount > int16(chair.RecoveryHP) {
+		return
 	}
+
+	if time.Since(plr.lastChairHeal) < 10*time.Second {
+		return
+	}
+
+	plr.giveHP(healAmount)
```
trameathia, line 582: is this actually needed? i swear when i sat in my chair it was already healing me correctly

ErwinsExpertise: The client automatically shows the HP icon above you, but doesn't actually grant any HP.

channel/field.go
```diff
@@ -549,6 +550,8 @@ type fieldInstance struct {
 	boatType   byte
 	properties map[string]interface{} // this is used to share state between npc and system scripts
 
+	chairs map[int32]int16
```
Hucaru: Redundant map? I can't see where it's used or any reason to store the chair usage in the instance?

channel/handlers.go
```diff
@@ -507,16 +525,54 @@ func (server Server) playerPassiveRegen(conn mnet.Client, reader mpacket.Reader)
...
 func (server Server) playerUseChair(conn mnet.Client, reader mpacket.Reader) {
-	fmt.Println("use chair:", reader)
-	// chairID := reader.ReadInt32()
+	plr, err := server.players.getFromConn(conn)
+	if err != nil {
+		log.Println(err)
+		return
+	}
+	chairID := reader.ReadInt32()
+	plr.chairSit(chairID)
```
Hucaru: Does this need to be a plr function? It's only called here and I can't think of any gameplay reason where the player is forced to sit without their input? I ideally prefere keeping as much of the logic together in the handler if possible.

channel/handlers.go
```diff
@@ -507,16 +525,54 @@ func (server Server) playerPassiveRegen(conn mnet.Client, reader mpacket.Reader)
...
-	if reader.ReadInt16() == -1 {
+	plr, err := server.players.getFromConn(conn)
+	if err != nil {
+		log.Println(err)
+		return
+	}
 
-	} else {
+	seatID := reader.ReadInt16()
+	plr.chairResult(seatID)
```
Hucaru: Ditto `plr.charSit`

channel/player.go
```diff
@@ -2065,6 +2068,69 @@ func (p *Player) updatePet() {
...
+		it := p.equip[i]
+		if it.slotID < 0 && it.amount > 0 && it.ID == itemID {
+			return true
+		}
+	}
+	return false
+}
+func (p *Player) chairSit(chairID int32) {
+	if chairID <= 0 {
+		delete(p.inst.chairs, p.ID)
```
Hucaru: Not needed? If removed do we need this if statement and the function could be embedded in the handler as
```
	p.chairID = reader.ReadInt32()
	p.inst.send(packetPlayerShowChair(p.ID, p.charID))
	p.Send(packetPlayerChairUpdate())
```

Does the chairID need to be stored as -1? Can't we just use the value the client sends?

channel/player.go
```diff
@@ -2065,6 +2068,69 @@ func (p *Player) updatePet() {
...
+
+	p.chairID = chairID
+	p.inst.send(packetPlayerShowChair(p.ID, chairID))
+	p.Send(packetPlayerChairUpdate())
+}
+
+func (p *Player) chairResult(chairID int16) {
+	if chairID == -1 {
+		if p.chairID != -1 {
+			delete(p.inst.chairs, p.ID)
```
Hucaru: Same comment as function above

channel/handlers.go
```diff
@@ -507,16 +525,62 @@ func (server Server) playerPassiveRegen(conn mnet.Client, reader mpacket.Reader)
...
+	plr, err := server.players.getFromConn(conn)
+	if err != nil {
+		log.Println(err)
+		return
+	}
+
+	chairID := reader.ReadInt32()
+	plr.chairID = chairID
+
+	plr.inst.send(packetPlayerShowChair(plr.ID, chairID))
```
Hucaru: Should use and stand skip active plr via `sendExcept(packetPlayerShowChair(plr.ID, chairID), plr.Conn)`?

PR #138, opened 2025-10-12
adding beginner skills

### LINKED ISSUE #133: beginner skills nimble feet and recovery don't actually work
when you use the skills they do their effects but nothing actually happens

trameathia:
fixes #133

### DIFF
--- a/channel/handlers.go
+++ b/channel/handlers.go
@@ -3931,10 +3931,9 @@ func (server *Server) playerSpecialSkill(conn mnet.Client, reader mpacket.Reader
 			go func(sid int32, cooldownTime int16, inst *fieldInstance) {
 				time.Sleep(time.Duration(cooldownTime) * time.Second)
 				// Use dispatch pattern for thread safety
-				if inst != nil && inst.dispatch != nil {
+				if inst != nil && inst.dispatch != nil && plr != nil {
 					inst.dispatch <- func() {
 						plr.Send(packetPlayerSkillCooldown(sid, 0))
-						log.Printf("packetPlayerSkillCooldown: skillID=%d, time=0", sid)
 					}
 				}
 			}(skillID, skillData.CooldownTime, plr.inst)


channel/buffs.go
```diff
@@ -68,14 +68,17 @@ func AddSkillBuff(skillID int32, bits ...int) {
 func LoadBuffs() {
 	skillBuffBits = make(map[int32][]int)
 
-	AddSkillBuff(int32(skill.NimbleBody), BuffSpeed)
+	// Beginner Skills
+	AddSkillBuff(int32(skill.NimbleFeet), BuffSpeed)
+	AddSkillBuff(int32(skill.Recovery), BuffHands) // Recovery buff (HP regen handled by 5-second ticker)
```
ErwinsExpertise, line 73: There should be an actual mask for this, but no idea what it is. Can we make sure this isn't actually buffing hands? If it is, you could always use 0 too.

channel/buffs.go
```diff
@@ -882,3 +900,50 @@ func (cb *CharacterBuffs) despawnSummonIfMatches(skillID int32) {
...
+
+				// Get skill data to retrieve the X value (recovery amount)
+				skillData, err := nx.GetPlayerSkill(int32(skill.Recovery))
+				if err != nil || currentLevel == 0 || int(currentLevel) > len(skillData) {
+					return
+				}
+
+				// Get the recovery amount from the X field
+				recoveryAmount := int16(skillData[currentLevel-1].X)
+				log.Printf("Recovery tick - Level: %d, X value: %d, HP before: %d/%d", currentLevel, recoveryAmount, cb.plr.hp, cb.plr.maxHP)
```
Hucaru: We probably don't want to print this out as it will polute logs

channel/buffs.go
```diff
@@ -882,3 +900,50 @@ func (cb *CharacterBuffs) despawnSummonIfMatches(skillID int32) {
...
+				if recoveryAmount > 0 {
+					cb.plr.giveHP(recoveryAmount)
+					log.Printf("HP after: %d/%d", cb.plr.hp, cb.plr.maxHP)
```
Hucaru: Same as above

channel/buffs.go
Hucaru: Has this been resolved?

trameathia: 0 would be buff weapon? from what i've seen nothing uses buff hands

channel/handlers.go
```diff
@@ -3958,6 +3958,26 @@ func (server *Server) playerSpecialSkill(conn mnet.Client, reader mpacket.Reader
...
+		// Send cooldown packet for beginner skills
+		if skillData, ok := plr.skills[skillID]; ok {
+			plr.Send(packetPlayerSkillCooldown(skillID, skillData.CooldownTime))
+			// Start timer to clear the cooldown when it expires
+			go func(sid int32, cooldownTime int16, inst *fieldInstance) {
+				time.Sleep(time.Duration(cooldownTime) * time.Second)
+				// Use dispatch pattern for thread safety
+				if inst != nil && inst.dispatch != nil {
+					inst.dispatch <- func() {
+						plr.Send(packetPlayerSkillCooldown(sid, 0))
```
Hucaru, line 3936: You need to check for `plr != nil` as they could have logged out

channel/handlers.go
```diff
@@ -3958,6 +3958,26 @@ func (server *Server) playerSpecialSkill(conn mnet.Client, reader mpacket.Reader
...
+						log.Printf("packetPlayerSkillCooldown: skillID=%d, time=0", sid)
```
Hucaru: We don't want to spam logs with skill usage information

ErwinsExpertise:
Hey @Hucaru I resolved those changes for @trameathia

PR #143, opened 2025-10-15
Add mob skills and debuffs

### LINKED ISSUE #104: Mob Skills
Attack skills for monsters work, however, we need to implement de/buff's and summons

ErwinsExpertise:
This PR will implement mob skills/debuffs.

Fixes #104

### DIFF
--- a/channel/player.go
+++ b/channel/player.go
@@ -825,15 +825,14 @@ func (d *Player) takeItem(id int32, slot int16, amount int16, invID byte) (Item,
 	}
 
 	item.amount -= amount
-	if item.amount == 0 {
-		// Delete item
+	if item.amount == 0 && !item.isRechargeable() {
 		d.removeItem(item)
 	} else {
-		// Update item with new stack size
 		d.updateItemStack(item)
 	}
 
 	return item, nil
+
 }
 
 func (d Player) updateItemStack(item Item) {
@@ -1171,7 +1170,7 @@ func (d *Player) consumeItemsByID(itemID int32, reqCount int32) bool {
 			if it.ID != itemID || it.amount <= 0 {
 				continue
 			}
-			take := int16(it.amount)
+			take := it.amount
 			if int32(take) > remaining {
 				take = int16(remaining)
 			}
--- a/channel/pools.go
+++ b/channel/pools.go
@@ -238,12 +238,13 @@ func (pool *lifePool) mobAcknowledge(poolID int32, plr *Player, moveID int16, sk
 			skillDelay := int16(skillData >> 16)
 
 			mobSkillID, mobSkillLevel, mobSkillData := pool.mobs[i].getMobSkill(skillDelay, skillLevel, skillID)
+
 			if mobSkillID != 0 {
 				pool.performSkill(mob, mobSkillID, mobSkillLevel, mobSkillData)
 			}
 
 			// Choose next skill
-			skillID, skillLevel = mob.canUseSkill(skillPossible)
+			skillID, skillLevel = mob.chooseNextSkill()
 
 			if !moveData.validateMob(v) {
 				return
@@ -256,6 +257,10 @@ func (pool *lifePool) mobAcknowledge(poolID int32, plr *Player, moveID int16, sk
 }
 
 func (pool *lifePool) performSkill(mob *monster, skillID, skillLevel byte, skillData nx.MobSkill) {
+	currentTime := time.Now().Unix()
+	mob.lastSkillTime = currentTime
+	mob.skillTimes[skillID] = currentTime
+
 	switch skillID {
 	case skill.Mob.Dispel:
 		for _, plr := range pool.instance.players {
--- a/channel/commands.go
+++ b/channel/commands.go
@@ -1137,18 +1137,6 @@ func (server *Server) gmCommand(conn mnet.Client, msg string) {
 	case "killAll":
 		fallthrough
 	case "killmobs":
-		var deathType byte = 1
-		if len(command) > 1 {
-			val, err := strconv.Atoi(command[1])
-
-			if err != nil {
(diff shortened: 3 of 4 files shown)


channel/player.go
```diff
@@ -1585,6 +1585,14 @@ func (d *Player) addForeignBuff(charId, skillID int32, level byte, delay int16)
 	d.buffs.AddBuff(charId, skillID, level, true, delay)
 }
 
+func (d *Player) addMobDebuff(skillID, level byte, durationSec int16) {
+	if d == nil || d.buffs == nil {
+		return
+	}
+	d.buffs.plr.inst = d.inst
```
Hucaru: Isn't this taking the player instance and setting it to the same value?

channel/buffs.go
```diff
@@ -729,6 +915,28 @@ func (cb *CharacterBuffs) ClearBuff(skillID int32, _ uint32) {
...
+		// Remove all positive skill IDs (player buffs)
+		// Mob debuffs use skill IDs 120-126, so we check if it's a typical player skill
+		if skillID >= 1000 {
+			toRemove = append(toRemove, skillID)
+		}
+	}
+	
+	// Remove each buff
+	for _, skillID := range toRemove {
+		cb.expireBuffNow(skillID)
```
Hucaru: Can this not be called in the previous loop instead of appending to a slice and then iterating the slice later?

channel/player.go
ErwinsExpertise: Yeah, not totally sure why it happens, but occasionally the pointer seems to go stale or something along those lines. This line just re-establishes the reference to make sure it’s consistent.

Hucaru: That's concerning!

I think it might be better if instead of holding a ptr to a plr in the buff struct the functions just take a plr ptr that way we don't need to be concerned about how stale things are?

So the calling code would look like:
```
func (d *Player) addMobDebuff(skillID, level byte, durationSec int16) {
	if d == nil || d.buffs == nil {
		return
	}

	d.buffs.AddMobDebuff(skillID, level, durationSec, d)
}
```

ErwinsExpertise: Fixed this weird stale thing and removed the assignment :)

channel/buffs.go
```diff
@@ -165,8 +165,8 @@ type CharacterBuffs struct {
 	expireAt          map[int32]int64  // sourceID -> unix ms expiry
 }
 
-func NewCharacterBuffs(p *Player) *CharacterBuffs {
-	return &CharacterBuffs{
```
Hucaru, line 169: Is this the root cause of the pointer going stale? If so I am not sure why it would be the case.

channel/buffs.go
Hucaru: I don't think this has been resolved? The extra iteration could still be removed

channel/pools.go
```diff
@@ -285,6 +290,70 @@ func (pool *lifePool) mobAcknowledge(poolID int32, plr *Player, moveID int16, sk
...
+	case skill.Mob.WeaponImmunity:
+		mob.statBuff |= skill.MobStat.PhysicalImmune
+		return
+	case skill.Mob.MagicImmunity:
+		mob.statBuff |= skill.MobStat.MagicImmune
+		return
+	}
+
+	// Get all players in the field instance
+	for _, plr := range pool.instance.players {
```
Hucaru: If you set this as the default condition of the switch statement you can then remove all the early returns

channel/buffs.go
ErwinsExpertise: Not 100% sure why it happens, but switching to mutate the existing player fixed it. My guess is the pointer inside buffs was getting set before the player was fully initialized, so sometimes it ended up pointing at an outdated reference. This way it always uses the final player instance and doesn’t go stale.

channel/buffs.go
ErwinsExpertise: Sorry, mixed it up. It's resolved now.

channel/pools.go
```diff
@@ -419,6 +441,72 @@ func (pool *lifePool) mobDamaged(poolID int32, damager *Player, dmg ...int32) {
...
+			summoner.spawnID,
+		)
+
+		if err != nil {
+			continue
+		}
+
+		var spawnedMob *monster
+		var highestID int32
+		for id, m := range pool.mobs {
```
Hucaru: If `spawnMobFromID` is modified to return `(error, *monster)` that would save having to iterate all monsters in the pool

https://github.com/Hucaru/Valhalla/blob/master/channel/pools.go#L503

channel/pools.go
```diff
@@ -419,6 +441,72 @@ func (pool *lifePool) mobDamaged(poolID int32, damager *Player, dmg ...int32) {
...
+			if i%2 == 1 {
+				offset = int16(35 * (i + 1) / 2)
+			} else {
+				offset = int16(-40 * (i / 2))
+			}
+		}
+		summonX := randX + offset
+		spawnPos := newPos(summonX, summonY, summoner.pos.foothold)
+
+		err := pool.spawnMobFromID(
```
Hucaru, line 481: This function sends a mob show packet so setting the summon properties further down won't affect the spawn.

https://github.com/Hucaru/Valhalla/blob/master/channel/pools.go#L486C21-L486C34

https://github.com/Hucaru/Valhalla/blob/master/channel/monster.go#L279-L281

channel/pools.go
ErwinsExpertise: I like the thought, but it doesn't make it any more dynamic. Instead I adjusted it to take summonType, although it seems the value that was being used was incorrect anyways, but still fixed this to be dynamic instead. Now we instead send spawnMobFromID the summonType, so if it's needed in the future we can easily do it.

Issue #94, opened 2025-09-04
Monsters Vanish During Map Transitions via Portals


LuneArk:
When entering a portal, the screen fades to black as the character transitions to the next map as it should. But during this transition, the monsters on the map vanish into thin air.

Hucaru:
I think this is a consiequence of removing the player from the life pool is the same function for when changing instance and using a portal. For the instance change we have to tell the client to unload the monsters where I suspect the client automatically unloads monsters when fading to black so we send a redundant packet causing the unload before fade to black. This line is the culprit I suspect:
https://github.com/Hucaru/Valhalla/blob/master/channel/pools.go#L213

PR #144, opened 2025-10-28
Fixed deadlock from special skills with zero cooldown, fixed field change dissapearing mobs before fade to black

Hucaru:
Fixed bugs:
- Special skills with zero cooldown e.g GM teleport cause a deadlock
- When changing field each mob would be removed from client via a packet causing mobs to dissapear before fade to black. This is not needed (unless changing instance in the same field) as the client auto unloads pools. #94

### DIFF
--- a/channel/buffs.go
+++ b/channel/buffs.go
@@ -993,11 +993,16 @@ func buildBuffMask(skillID int32) *Flag {
 }
 
 func (cb *CharacterBuffs) post(fn func()) {
-	if cb.plr != nil && cb.plr.inst != nil && cb.plr.inst.dispatch != nil {
-		cb.plr.inst.dispatch <- fn
+	if cb == nil || cb.plr == nil || cb.plr.inst == nil || cb.plr.inst.dispatch == nil {
 		return
 	}
-	fn()
+	select {
+	case cb.plr.inst.dispatch <- fn:
+		return
+	default:
+		fn()
+		return
+	}
 }
 
 func (cb *CharacterBuffs) scheduleExpiryLocked(skillID int32, after time.Duration) {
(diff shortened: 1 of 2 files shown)


channel/buffs.go
```diff
@@ -1008,7 +1008,7 @@ func (cb *CharacterBuffs) scheduleExpiryLocked(skillID int32, after time.Duratio
 	}
 
 	if after <= 0 {
-		cb.post(func() { cb.expireBuffNow(skillID) })
+		cb.expireBuffNow(skillID)
```
Hucaru: @ErwinsExpertise can you forsee any issue with this change? 

Previously on a skill like gm teleport (zero cooldown) it creates a teleport buff icon that never dissappears and deadlocks the dispatch go routine which causes the server to no longer shut down as it cannot flush players.

ErwinsExpertise: Should be fine, but I'd be more inclined to adjust post instead. 

What are your thoughts on this?
```
func (cb *CharacterBuffs) post(fn func()) {
	if cb == nil || cb.plr == nil || cb.plr.inst == nil || cb.plr.inst.dispatch == nil {
		return
	}
	select {
	case cb.plr.inst.dispatch <- fn:
		return
	default:
		fn()
		return
	}
}
```

Also, I think how `useSkill` is being handled is also preventing this from pushing forward properly because it's a GM skill. 

https://github.com/Hucaru/Valhalla/blob/master/channel/handlers.go#L4117

should probably be changed to something like
```
	plr.useSkill(skillID, skillLevel, 0)
	plr.Send(packetPlayerNoChange())
```

We should actually probably move useSkill to the top of `playerSpecialSkill` 🙃 (oops). I think it's always a good idea to send a nochange packet no matter what to make sure the client can continue.

PR #151, opened 2025-10-30
fix warpPlayer function for spawn portal

### LINKED ISSUE #150: When warping into map via login or warp command it is not random
When using the `/warp` command or character login the portal is not random.

Hucaru:
Closes #150

### DIFF
--- a/channel/handlers.go
+++ b/channel/handlers.go
@@ -3909,6 +3909,25 @@ func (server *Server) playerSpecialSkill(conn mnet.Client, reader mpacket.Reader
 			}
 		}
 
+	// Nimble feet and recovery beginner skills with cooldown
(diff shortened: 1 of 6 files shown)


channel/handlers.go
```diff
@@ -1025,20 +1025,9 @@ func (server Server) warpPlayer(plr *Player, dstField *field, dstPortal portal,
 
 	plr.setMapID(dstField.id)
 	plr.pos = dstPortal.pos
-	var spawnIdx byte
 
-	dPortal, err := dstInst.getPortalFromName(dstPortal.name)
```
Hucaru, line 1030: This always returns the first portal as spawn portals all have the name "sp"

ErwinsExpertise: Makes sense

PR #157, opened 2025-11-01
Changed how players are stored & fixed world server not sending cashshop channel info

### LINKED ISSUE #158: Cash shop cannot send player back to channel
The cash shop does not have channel information and fails to send player back to channel

Hucaru:
- Changed how servers store players from slice to maps with conn, name, id keys pointing to same players
  - This should also provide a speed increase when player counts are large due to constant time lookup
- Fixed world server to send cash shop channel ip/port information otherwise player cannot rejoin channel. Fixes #158
- Moved portal list fix into nx parsing

@ErwinsExpertise I would appreciate it if you could pull this branch and test it in case I have missed some edge case.


channel/handlers.go
```diff
@@ -3405,20 +3403,17 @@ func (server *Server) handlePlayerConnectedNotifications(conn mnet.Server, reade
 		}
 	}
 
-	for i, v := range server.players {
-		if v.ID == playerID {
-			continue
-		} else if v.hasBuddy(playerID) {
+	server.players.observe(func(plr *Player) {
```
ErwinsExpertise: This appears to iterate over _all_ of the players in the channel to identify buddies and sends the packets to the targets. Would it not be better to simply iterate over the players buddy list and if any are online send it to them? A buddy list is generally ~30 characters while the active player count in a channel could be much larger. I suppose ideally, the world server would handle this logic by looking at the players buddies, identify which are online & what channel they are on, and send the packets specifically to the appropriate channels.

ErwinsExpertise (approved):
@Hucaru this looks good, and I definitely like this **a lot** more than using slices to hold the player structs. Awesome refactor. 

Also, I pulled the branch and tested quite a bit and did not notice any regressions, or bugs.

channel/handlers.go
Hucaru: So this is the handler for when the world server reports to channel servers that a player of id with name, guild id etc has loged in. We therefore have to iterate over all players as only 1 channel will have the player.

EDIT: I think this has highlited that the handlers for client and server should be split into different files for the channel server as it's gotten too long

channel/handlers_server.go
Hucaru, line 1: @ErwinsExpertise are you in agreement with spliting the handlers into two files for server and client?

ErwinsExpertise: Yes that makes complete sense to me. It is a little confusing looking between both so this should help a lot.

channel/handlers.go
ErwinsExpertise: Ah ok. I've wondered if using a map for channels might be better, because then we could assign the channel to what it's supposed to be and target channels specifically. Right now channels are all in a slice, which is fine, but it makes it so it's harder to target a channel. If they were in a map[int]*Channel we could then just look at what channel # the player's buddies(via DB) are in and send the login packets to them specifically. 

Currently, the world server tells the channel what #(id) it should be:
https://github.com/Hucaru/Valhalla/blob/master/world/handlers.go#L82

That could be changed to where when a channel registers it tells the world what channel it wants to register as, and if 0/null then fallback to 1...20 .

Hucaru: Not sure I completely follow. The world server doesn't keep track of players so it wouldn't know what channels to send player login event messages to. I will get the current changes merged in. If you want, raise an issue with your suggested change and we can discuss there.

PR #162, opened 2025-11-01
Removed portal scripts, moved into handler directly with unified warp function

Hucaru:
There are only a fixed small number of portals and their functionality will never change, therefore:
- Removed portal scripts and associated script manager
- The scripted portals all follow the same warp process so unified them into one function call with a switch statement

Events will also be moved into an event manager for boats and bosses as they are static fixtures


ErwinsExpertise (approved):
Looks great!

PR #167, opened 2025-11-03
Better boats & boss monitoring

Hucaru:
- Boat management is now handled in go and is much more accurate and perfomant than previously
- Added clock time to stations
- Boss spawn from reactor interaction is now managed in go
- Removed event js script system

### DIFF
--- a/channel/boss.go
+++ b/channel/boss.go
@@ -44,10 +44,10 @@ func manageSummonedBoss(inst *fieldInstance, mobID int32, server *Server) {
 		case <-timeout.C:
 			var returnMap int32
 
-			switch mobID {
-			case constant.MobPapalatusBall:
+			switch inst.fieldID {
+			case constant.MapBossPapulatus:
 				returnMap = constant.MapBossPapulatusReturn
-			case constant.MobZakum1Body:
+			case constant.MapBossZakum:
 				returnMap = constant.MapBossZakumReturn
 			}
 
@@ -78,6 +78,7 @@ func manageSummonedBoss(inst *fieldInstance, mobID int32, server *Server) {
 				}
 
 				inst.lifePool.eraseMobs()
+				inst.reactorPool.reset(false)
 				inst.properties["eventActive"] = false
 			}
 			return
--- a/channel/commands.go
+++ b/channel/commands.go
@@ -1694,5 +1694,5 @@ func covnertMobNameToID(name string) ([]int32, error) {
 		return []int32{constant.MobZombieMushmom}, nil
 	}
 
-	return nil, fmt.Errorf("unkown mob Name")
+	return nil, fmt.Errorf("unknown mob Name")
 }
--- a/channel/pools.go
+++ b/channel/pools.go
@@ -1164,7 +1164,7 @@ func (pool *reactorPool) playerShowReactors(plr *Player) {
 	}
 }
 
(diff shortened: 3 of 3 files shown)


channel/boats.go
```diff
@@ -0,0 +1,206 @@
...
+	r := rand.New(rand.NewSource(time.Now().Unix()))
+
+	server.dispatch <- func() {
+		boatsBoarding(server, true)
+	}
+
+	// Boats take off and land every 15 mins from the hour, with boarding starting 5 mins before
+	for {
+		now := time.Now()
+		delay := roundUpTime(now, 15*time.Minute).Sub(now)
```
ErwinsExpertise, line 37: Are we able to make the clock/timer in-game at the station reflect the appropriate times?

channel/boats.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+
+		for _, inst := range field.instances {
+			if finish {
+				inst.lifePool.eraseMobs()
+			}
+
+			if len(inst.lifePool.mobs) == 0 || finish {
+				inst.showBoats(false, 0x01)
+				inst.changeBgm("Bgm04/UponTheSky") // TODO: Find the correct bgm
```
ErwinsExpertise: This is the correct bgm for the map

channel/boats.go
Hucaru: I'm not sure what does this. I believe it's a clock so I suspect it might be part of character enter channel?

Hucaru: Added

channel/boss.go
```diff
@@ -0,0 +1,97 @@
...
+					}
+				}
+
+				inst.properties["eventActive"] = false
+				finished.Store(true)
+			}
+		case <-timeout.C:
+			var returnMap int32
+
+			switch mobID {
```
ErwinsExpertise: Seems odd to switch based on mob id, why not use the map instead? If we use map ID it is constant and will not be subject to possible race conditions with the timer.

channel/boss.go
```diff
@@ -0,0 +1,97 @@
...
+				if err != nil {
+					log.Println(err)
+					return
+				}
+
+				for _, plr := range inst.players {
+					server.warpPlayer(plr, field, portal, true)
+				}
+
+				inst.lifePool.eraseMobs()
```
ErwinsExpertise, line 80: We need to reset the reactors too:
https://github.com/Hucaru/Valhalla/blob/master/channel/pools.go#L1165

channel/commands.go
```diff
@@ -1622,30 +1669,30 @@ func convertJobNameToID(name string) int16 {
...
 	case "mushmom":
-		return []int32{6130101}, nil
+		return []int32{constant.MobMushmom}, nil
 	case "zmushmom":
-		return []int32{6300005}, nil
+		return []int32{constant.MobZombieMushmom}, nil
 	}
 
-	return nil, fmt.Errorf("Unkown mob Name")
+	return nil, fmt.Errorf("unkown mob Name")
```
ErwinsExpertise: typo: "unknown"

channel/boss.go
Hucaru: Chnaged as map is less error prone. Not sure I follow the timer argument

ErwinsExpertise (approved):
Looks good!

PR #170, opened 2025-11-05
Resolve boss not cleaning up properly

ErwinsExpertise:
This pull request updates the boss management logic in `channel/boss.go` to improve cleanup and event handling. The main changes ensure that mobs, drops, and reactors are properly reset when there are no players present, which helps prevent lingering state and potential bugs.

Boss event cleanup improvements:

* Added checks to only perform boss cleanup (erasing mobs, drops, and resetting reactors) if there are no players in the instance, preventing premature cleanup while players are still present.
* Ensured that drops are also erased during boss cleanup, in addition to mobs and reactors, to fully reset the event state.

### DIFF
--- a/channel/boss.go
+++ b/channel/boss.go
@@ -32,19 +32,13 @@ func manageSummonedBoss(inst *fieldInstance, mobID int32, server *Server) {
 			}
 
 			inst.dispatch <- func() {
-				if len(inst.players) > 0 {
-					for _, mob := range inst.lifePool.mobs {
-						if mob.boss {
-							return
-						}
-					}
+				if len(inst.players) <= 0 {
+					inst.lifePool.eraseMobs()
+					inst.dropPool.eraseDrops()
+					inst.reactorPool.reset(false)
+					inst.properties["eventActive"] = false
+					finished.Store(true)
 				}
-
-				inst.lifePool.eraseMobs()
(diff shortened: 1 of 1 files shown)


channel/boss.go
```diff
@@ -32,12 +32,17 @@ func manageSummonedBoss(inst *fieldInstance, mobID int32, server *Server) {
...
+				if len(inst.players) > 0 {
+					for _, mob := range inst.lifePool.mobs {
+						if mob.boss {
+							return
+						}
 					}
 				}
 
+				inst.lifePool.eraseMobs()
+				inst.dropPool.eraseDrops()
```
Hucaru: This will erase drops from mob potentially before player picks them up after killing the boss?

channel/boss.go
```diff
@@ -32,12 +32,17 @@ func manageSummonedBoss(inst *fieldInstance, mobID int32, server *Server) {
...
-						return
+				if len(inst.players) > 0 {
+					for _, mob := range inst.lifePool.mobs {
+						if mob.boss {
+							return
+						}
 					}
 				}
 
+				inst.lifePool.eraseMobs()
```
Hucaru: I think original behaviour is: spawned mobs remain in field even if boss has died

Issue #166, opened 2025-11-03
Implement monster sacks, boss spawn items


Hucaru:
Currently the following items don't work as intended:

<img width="182" height="294" alt="Image" src="https://github.com/user-attachments/assets/0e3eec90-3322-420e-96d1-ac4f7c6a5f44" />

<img width="176" height="291" alt="Image" src="https://github.com/user-attachments/assets/deeab005-10d4-4061-8835-52b066e1686c" />

PR #171, opened 2025-11-05
Sacks are consumed

Hucaru:
Implements #166

### DIFF
--- a/channel/boss.go
+++ b/channel/boss.go
@@ -32,14 +32,13 @@ func manageSummonedBoss(inst *fieldInstance, mobID int32, server *Server) {
 			}
 
 			inst.dispatch <- func() {
-				for _, mob := range inst.lifePool.mobs {
-					if mob.boss {
-						return
-					}
+				if len(inst.players) <= 0 {
+					inst.lifePool.eraseMobs()
+					inst.dropPool.eraseDrops()
+					inst.reactorPool.reset(false)
+					inst.properties["eventActive"] = false
+					finished.Store(true)
 				}
-
-				inst.properties["eventActive"] = false
-				finished.Store(true)
 			}
 		case <-timeout.C:
 			var returnMap int32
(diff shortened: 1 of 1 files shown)


channel/player.go
```diff
@@ -1088,7 +1088,24 @@ func (d *Player) moveItem(start, end, amount int16, invID byte) error {
 			return fmt.Errorf("unable to take Item")
 		}
 
-		d.inst.dropPool.createDrop(dropSpawnNormal, dropFreeForAll, 0, d.pos, true, d.ID, 0, dropItem)
+		if len(dropItem.spawnMobs) > 0 {
+			d.inst.dropPool.createDrop(dropSpawnDisappears, dropSpawnDisappears, 0, d.pos, true, d.ID, 0, dropItem)
```
Hucaru: Not too sure what these drop types should be, I suspect it should be the one where it disappears at the top of the drop instead of on the landing?

channel/player.go
```diff
@@ -1088,7 +1088,24 @@ func (d *Player) moveItem(start, end, amount int16, invID byte) error {
...
+
+			for mobID, prob := range item.spawnMobs {
+				if prob >= int32(d.randIntn(100)) {
+					summonType := constant.MobSummonTypePoof
+
+					switch mobID {
```
Hucaru: Not sure if there are any other summonTypes that are applicable here

ErwinsExpertise:
I’m fairly confident this is a consume item? The spawn should trigger when you use it rather than drop.

ErwinsExpertise:
> > I’m fairly confident this is a consume item? The spawn should trigger when you use it rather than drop.
> 
> Yes, it's a consume item but I remember it being a visible drop? Or am I just miss remembering?
> 
> EDIT: I must have been miss remembering as consuming sacks has it's own op code. Any idea what this is? It's not interactable as far as I can tell
> 
> image

That looks like part of Korean Folk Town. Just extra stuff for later versions Nexon shoved in wz.

PR #184, opened 2025-11-09
Add Forced Return

ErwinsExpertise:
This pull request introduces several enhancements and bug fixes related to skill usage restrictions, map handling, and network packet management. The most significant change is the implementation of a check that prevents players from using skills while under a "Seal" mob debuff. Additionally, there are improvements to map forced return handling and new network opcodes for skill state communication.

**Skill usage and debuff handling:**
* Added a `hasMobDebuff` method to `CharacterBuffs` to check for active mob debuffs by skill ID. This is used to determine if a character is currently affected by a debuff such as "Seal".
* Updated the `useSkill` method in `Player` to prevent skill usage when the "Seal" debuff is active, returning an appropriate error.

**Map forced return handling:**
* Changed the type of `ForcedReturn` in the `Map` struct from `int64` to `int32` for consistency and correctness.
* Updated the map info loader to parse `forcedReturn` as `int32` instead of `int64`.
* Enhanced the `Logout` method in `Player` to check for a forced return map on logout and update the player's map accordingly.

**Network packet and opcode updates:**
* Added new opcodes `SendChannelPlayerPrepareSkill` and `SendChannelPlayerStopSkill` to support communication about skill preparation and interruption.
* Implemented a new `packetSkillStop` function to create packets for stopping a player's skill, using the new opcode.

**Minor change:**
* Commented out a redundant call to `sendChannelInfo` in the cash shop handler to avoid unnecessary updates.

### DIFF
--- a/channel/handlers_client.go
+++ b/channel/handlers_client.go
@@ -1051,10 +1051,6 @@ func (server Server) warpPlayer(plr *Player, dstField *field, dstPortal portal, 
 
 	plr.setMapID(dstField.id)
 
-	if plr.party != nil {
-		plr.UpdatePartyInfo(plr.party.ID, plr.ID, int32(plr.job), int32(plr.level), dstField.id, plr.Name)
-	}
-
 	plr.pos = dstPortal.pos
 
 	plr.Send(packetMapChange(dstField.id, int32(server.id), dstPortal.id, plr.hp))


world/handlers.go
```diff
@@ -169,7 +169,7 @@ func (server *Server) handleNewCashShop(conn mnet.Server, reader mpacket.Reader)
 	log.Println("Registered CashShop")
 
 	server.sendCashShopInfo()
-	server.sendChannelInfo() // Slightly waistfull as it updates all the channels as well
+	// server.sendChannelInfo() // Slightly waistfull as it updates all the channels as well
```
Hucaru: Why disable this?

channel/player.go
```diff
@@ -1316,6 +1320,16 @@ func (d Player) Logout() {
 		}
 	}
 
+	currentMap, err := nx.GetMap(d.mapID)
```
Hucaru: I wonder if doing this on player connect is a better location. If something goes wrong during server shutdown then players would spawn back in the wrong map?

world/handlers.go
ErwinsExpertise: Accident! I was looking into something let me remove that

channel/player.go
ErwinsExpertise: 🤔 yeah I get what you are saying. I can change it to that

ErwinsExpertise:
@Hucaru I moved the logic into the `warpPlayer` function, because the only way the scenario would happen is during warp events as players cannot spawn in those maps. This should address your shutdown concerns.

channel/handlers_client.go
```diff
@@ -1039,7 +1039,17 @@ func (server Server) warpPlayer(plr *Player, dstField *field, dstPortal portal,
...
+	if err != nil {
+		log.Println(err)
+	}
+
+	if dstMap.ForcedReturn != 999999999 {
+		plr.setMapID(dstMap.ForcedReturn)
+	} else {
+		plr.setMapID(dstField.id)
+	}
+
```
Hucaru, line 1053: Won't this force people who are using a portal to get sent to the forced return map? Also means GM's can't warp to the boss's/boats?

ErwinsExpertise: The actual packet itself is still sending them to the boss room, however, we are setting their mapID in DB to the return map.

Hucaru: I think it would be better to perform this check once on player connect instead of every warp interaction.

Doing it this way doesn't mark anything to need to be updated in the database and having the player map ID not be their actual map ID could have some unintended side effects as well.

ErwinsExpertise: That makes more sense :) Changed

channel/handlers_client.go
```diff
@@ -1039,16 +1049,7 @@ func (server Server) warpPlayer(plr *Player, dstField *field, dstPortal portal,
...
-	}
-
-	if dstMap.ForcedReturn != 999999999 {
-		plr.setMapID(dstMap.ForcedReturn)
-	} else {
-		plr.setMapID(dstField.id)
-	}
+	plr.setMapID(dstField.id)
 
 	if plr.party != nil {
```
Hucaru: I think this can be moved back into the player setMapID method

ErwinsExpertise: Oops, I did add it back, but forgot to remove it from here.

Hucaru (approved):

Issue #107, opened 2025-09-17
Zakum Quest Line


ErwinsExpertise:

ErwinsExpertise:
ref: https://github.com/sewil/OpenMG/blob/main/Scripts/zakum.s

ErwinsExpertise:
> I'm not familiar with the quest line but I assume you have to do it sequentially, with each stage giving rewards that lead up to the reactor summon item?

Yeah, I found this guide which seems accurate to GMS as far as I remember:
https://forum.maplelegends.com/index.php?threads/zakum-prequest-guide.2738/

All of these stages can be repeated as often as you want, but you need to do all 3 at least once to be able to get into zak from what I remember.

PR #194, opened 2025-11-11
Implement Mystic Doors

ErwinsExpertise:
This pull request introduces a comprehensive overhaul to the mystic door (town portal) system and related buff mechanics, improving both the underlying data structures and the handling of party and player interactions with doors and buffs. The changes include the addition of new GM skill buffs, a more robust management of mystic doors and portals, and enhancements to debuff curing functionality. The most important changes are grouped below by theme.

### Mystic Door System Overhaul

* Added a new `channel/door.go` file implementing all logic for creating, showing, and removing mystic doors and their associated portals, including party-specific handling and expiration.
* Updated `fieldInstance` to track mystic doors (`mysticDoors`), pending door syncs, and introduced the `mysticDoorInfo` struct for detailed door state. Methods for portal management, door display, and party synchronization were added. [[1]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR326-R327) [[2]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR551-R554) [[3]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR579-R602) [[4]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR696-R697) [[5]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR711-R770) [[6]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR896-R952) [[7]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR975-R1005)
* Refactored portal packet functions for proper encoding of mystic/town/party portals and their removal, ensuring correct client display and synchronization. [[1]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aL919-L928) [[2]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aL989-R1204)
* Added logic to expire mystic doors after their duration and synchronize party door bindings during field updates.

### Buff and Debuff Mechanics

* Replaced old GM skill buff registrations with new Super GM skill buffs, adding support for additional haste and holy symbol variants.
* Updated buff checks to use new skill IDs (e.g., `SuperGMHide`) and added a new method `HasHolySymbol` to handle both regular and GM holy symbol buffs, including party size-based multipliers.
* Added a method `cureAllDebuffs` to `CharacterBuffs`, allowing for mass removal of multiple debuff types in one call.

### Server and Handler Updates

* Updated client packet handler to support buff cancellation requests, mapping the new opcode to the appropriate server method.
* Fixed forced return map logic to use the constant for invalid maps, improving map transition reliability.

These changes collectively modernize and stabilize the mystic door and buff systems, providing better support for party play, GM features, and debuff management.

[[1]](diffhunk://#diff-279076f2c231870145d085552a98844632a8bec6bc9f17596a40a32c57b27d3bR1-R183) [[2]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR326-R327) [[3]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR551-R554) [[4]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR579-R602) [[5]](diffhunk://#diff-a6454022993b0895c14ed89b020f546bf952f4429fea2971610704be8cd2e230L145-R149) [[6]](diffhunk://#diff-a6454022993b0895c14ed89b020f546bf952f4429fea2971610704be8cd2e230L183-R223) [[7]](diffhunk://#diff-a6454022993b0895c14ed89b020f546bf952f4429fea2971610704be8cd2e230R579-R613) [[8]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR126-R127) [[9]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL244-R246)


channel/field.go
```diff
@@ -669,6 +688,71 @@ func (inst *fieldInstance) addPlayer(plr *Player) error {
...
+			isParty := false
+			ownerIdx := byte(0)
+			for i, pid := range plr.party.PlayerID {
+				if pid == ownerID {
+					isParty = true
+					ownerIdx = byte(i)
+					break
+				}
+			}
+			if isParty {
```
Hucaru: This if statement can be within the for loop removing the need for `isParty` bool

channel/field.go
```diff
@@ -794,6 +878,51 @@ func (inst fieldInstance) getPortalFromID(id byte) (portal, error) {
 	return portal{}, fmt.Errorf("No portal with that Name")
 }
 
+func (inst *fieldInstance) getNextPortalID() byte {
```
Hucaru, line 901: Currently portal ids are being shared accross parties. Can you not assign portal id's based off of `255 - player party index` and have it be part of the mystic door struct this way no calculation is needed for portals and no instance management is needed as only party members receive portal creation/deletion packets so it doesn't matter if different parties have the same portal id.

As far as I am aware no map comes anywhere close to having 255 portals?

channel/handlers_client.go
```diff
@@ -448,8 +448,115 @@ func (server Server) playerEmote(conn mnet.Client, reader mpacket.Reader) {
...
+		plr.Send(packetPlayerNoChange())
+		return
+	}
+
+	var (
+		nearestDoor *mysticDoorInfo
+		ownerID     int32
+		minDist     int32 = 1<<30 - 1
+	)
+	for oid, door := range plr.inst.mysticDoors {
```
Hucaru, line 468: Could attach mystic doors to players and then iterate over party players instead of the whole instance

channel/handlers_server.go
```diff
@@ -283,7 +283,15 @@ func (server *Server) handlePartyEvent(conn mnet.Server, reader mpacket.Reader)
 			plr, _ := server.players.GetFromID(playerID)
 			party.SerialisePacket(&reader)
 			party.addPlayer(plr, index)
+
+			if plr != nil && plr.inst != nil {
+				inst := plr.inst
+				inst.dispatch <- func() {
```
Hucaru: Why is this being dispatched? I can't see any timing related reason?

channel/player.go
```diff
@@ -255,6 +255,14 @@ type Player struct {
 	summons *summonState
 	pet     *pet
 
+	// Mystic Door tracking
+	doorMapID       int32
```
Hucaru, line 259: As mentioned above, you could encapsulate everything related to a door in a single struct and attach it to the plr.

channel/handlers_client.go
ErwinsExpertise: This is in reply to both this and https://github.com/Hucaru/Valhalla/pull/194#discussion_r2516078002

----

Traditionally, mystic doors are an instance-managed entity in GMS, and the current approach is simpler as it avoids complex player-based checks every time a person enters a map. This also ensures consistency, and is true to original GMS behavior where all players on a map could see a door even if they couldn't use it. While attaching doors to players might seem better for iteration, managing them at an instance level has minimal performance impact and aligns with existing map structure logic and the use of designated `tp` portals.

I've implemented the doors this way due to the simplicity and lack of significant drawbacks, but I understand you prefer the player-attached approach. I'm willing to implement the changes using player-based tracking and portal ID management if you can provide a bit more detail or guidance on how you'd like it structured.

channel/field.go
ErwinsExpertise: I appreciate the idea, but we need instance-unique portal IDs here. The client handles door spawn/remove alongside the map’s static portals, and the two sides of a door (source and town) use different packet flows. Reusing IDs per party risks collisions with existing “sp/tp” entries and breaks when party membership changes since the party index is not stable. We already gate usage via the party-binding packet, so party scoping is preserved without changing ID allocation. The current allocator is trivial and the instance-level tracking we keep is required anyway for town-door slot reservation, expiry, and cleanup. Given how much of this feature relies on town portals and cross-map behavior, keeping IDs unique per instance is the safer and simpler path.

channel/handlers_client.go
Hucaru: I think it's ok to keep it as is. I don't have super strong feeling either way.

channel/door.go
```diff
@@ -0,0 +1,227 @@
...
+		if returnField, ok := plr.inst.server.fields[returnMapID]; ok {
+			if returnInst, err := returnField.getInstance(0); err == nil {
+				createTownDoor(plr, returnInst, doorPos, expiresAt)
+			}
+		}
+	}
+}
+
+func createSourceDoor(plr *Player, doorPos pos, expiresAt time.Time) {
+	doorSpawnID := plr.inst.idCounter
```
Hucaru: there is a function for this `inst.nextID()`?

channel/door.go
```diff
@@ -0,0 +1,227 @@
...
+		townPortal:  false,
+		srcPos:      doorPos,
+		expiresAt:   expiresAt,
+	}
+
+	plr.inst.send(packetMapSpawnMysticDoor(doorSpawnID, doorPos, true))
+	plr.inst.send(packetMapPortal(plr.mapID, plr.inst.returnMapID, doorPos))
+
+	if plr.party != nil {
+		for _, viewer := range plr.inst.players {
```
Hucaru: Iterate over party members and check their map id's instead of iterating over the instance and checking their ids?

Also means you don't need to check party id matches or if they are in a party

I think the following does the same thing and is a repeated pattern in this file

```
for index, viewer := range plr.party.players {
  if viewer == nil {
    continue
  }

  viewer.Send(packetMapPortalParty(byte(index), plr.mapID, plr.inst.returnMapID, doorPos))
}
```

ErwinsExpertise:
> What happens if a player drops a door in a party and then disconnects and re-connects?

Door stays until it expires. IIRC that is true to original behavior as well. I don't recall doors disappearing when the caster leaves since they are tied to instance.

channel/door.go
```diff
@@ -0,0 +1,244 @@
...
+		viewer.Send(packetMapPortal(plr.mapID, townInst.fieldID, doorPos))
+	}
+
+	if plr.party != nil {
+		for _, viewer := range townInst.players {
+			if viewer == nil || viewer.party == nil || viewer.party.ID != plr.party.ID {
+				continue
+			}
+			ownerIdx := byte(0)
+			for i, pid := range viewer.party.PlayerID {
```
Hucaru: This for loop can be done once before instead on every iteration

channel/door.go
```diff
@@ -0,0 +1,244 @@
...
+					ownerIdx = byte(i)
+					break
+				}
+			}
+			viewer.Send(packetMapPortalParty(ownerIdx, plr.mapID, townInst.fieldID, doorPos))
+		}
+	}
+
+	plr.Send(packetMapPortal(plr.mapID, townInst.fieldID, doorPos))
+	if plr.party != nil {
```
Hucaru: This can be combined into above if?

channel/door.go
```diff
@@ -0,0 +1,244 @@
...
+				removeDoorFromInstance(doorInst, plr.doorSpawnID, plr.doorPortalIndex, plr.ID)
+			}
+		}
+	}
+
+	if plr.townDoorMapID != 0 {
+		if townField, ok := plr.inst.server.fields[plr.townDoorMapID]; ok {
+			if townInst, err := townField.getInstance(0); err == nil {
+				if plr.townPortalIndex >= 0 && plr.townPortalIndex < len(townInst.portals) {
+					townInst.portals[plr.townPortalIndex].destFieldID = constant.InvalidMap
```
Hucaru: We probably need an `instance.createPortal()` function that returns the portal created (as you might want it's name or id) since we do this in multiple locations and could also be re-used by things like gm commands

channel/door.go
```diff
@@ -0,0 +1,244 @@
...
+					townInst.portals[plr.townPortalIndex].name = "tp"
+					townInst.portals[plr.townPortalIndex].temporary = false
+				}
+				removeDoorFromInstance(townInst, plr.townDoorSpawnID, -1, plr.ID)
+			}
+		}
+	}
+
+	// Reset player-side door state
+	plr.doorMapID = 0
```
Hucaru: `plr.unsetMysticDoor()` can use in multiple places

ErwinsExpertise:
Added the basics for #107 , the main thing we are waiting on is the PQ logic from @Hucaru , but in a non-public server this will work just fine.

channel/party.go
```diff
@@ -46,6 +46,15 @@ func (d *party) addPlayer(plr *Player, index int32) {
 	d.broadcast(packetPartyPlayerJoin(d.ID, d.Name[index], d))
 }
 
+func (d *party) getPlayerID(plrID int32) byte {
```
Hucaru: This should be `getPlayerPartyIndex()` or something similar. It's technically returning an index not an ID and is too similar to the prexisting playerID which represents database ID

channel/field.go
```diff
@@ -794,6 +893,63 @@ func (inst fieldInstance) getPortalFromID(id byte) (portal, error) {
...
```
Hucaru, line 901: This is a relatively expensive function for what it's trying to do. I think a different approach for portal id re-use needs to be attempted. Maybe something similar to:

```
func (f *field) createInstance(rates *rates, server *Server) int {
	id := len(f.instances)

	inst := &fieldInstance{
		id:          id,
		fieldID:     f.id,
		portals:     portals,
		dispatch:    f.Dispatch,
		town:        f.Data.Town,
		returnMapID: f.Data.ReturnMap,
		timeLimit:   f.Data.TimeLimit,
		properties:  make(map[string]interface{}),
		fhHist:      f.fhHist,
		server:      server,
	}

	for i := 0; i < len(inst.portals); i++ {
        if i <= len(f.Data.Portals) {
            inst.portals[i] = createPortalFromData(f.Data.Portals[i])
        } else {
            inst.portals[i] = portal{id: byte(i), destFieldID: constant.InvalidMap}
        }		
	}
}

type fieldInstance struct {
    ...
	portals [256]portal
}
```

```
func (inst *fieldInstance) getNextPortalID() byte {
    for _, p := range inst.portals {
        if p.dstFieldID == constant.InvalidMap {
            return p.id
        }
    }

    return 255
}
```
Portal removal is then:
```
func (inst *fieldInstance) removePortalAtIndex(index int) {
	if index >= 0 && index < len(inst.portals) {
		inst.portals[index] = constant.InvalidMap
	}
}
```
Adding portal function could go as it's only called by createNewPortal I think and could be combined into:
```
func (inst *fieldInstance) createNewPortal(pos pos, name string, destFieldID int32, destName string, temp bool) int {
    id := inst.getNextPortalID()

	inst.portals[id] = portal{
		id:          id,
		name:        name,
		pos:         pos,
		destFieldID: destFieldID,
		destName:    destName,
		temporary:   temp,
    }

	return id
}
```

channel/handlers_server.go
```diff
@@ -650,6 +655,95 @@ func (server *Server) playerSpecialSkill(conn mnet.Client, reader mpacket.Reader
...
+			}
+		}
+
+	case skill.Heal:
+		plr.inst.send(packetPlayerSkillAnimation(plr.ID, false, skillID, skillLevel))
+
+		for _, member := range plr.party.players {
+			if member.mapID != plr.mapID {
+				continue
+			}
```
Hucaru: I believe heal has a party member range requirement?

channel/field.go
ErwinsExpertise: @Hucaru Agreed, just refactored.

channel/field.go
```diff
@@ -309,22 +309,32 @@ type field struct {
...
-		dispatch:    f.Dispatch,
-		town:        f.Data.Town,
-		returnMapID: f.Data.ReturnMap,
-		timeLimit:   f.Data.TimeLimit,
-		properties:  make(map[string]interface{}),
-		fhHist:      f.fhHist,
-		server:      server,
+		id:              id,
+		fieldID:         f.id,
+		portals:         portals,
```
Hucaru: This will generate a slice from the array. To prevent this the fieldInstance definition needs to be `portals: [256]portal`.

You can then move the above iterations to be after the `inst := ` statement and iterate over `len(inst.portals)`

channel/field.go
```diff
@@ -785,15 +892,67 @@ func (inst fieldInstance) getPortalFromName(name string) (portal, error) {
 }
 
 func (inst fieldInstance) getPortalFromID(id byte) (portal, error) {
-	for _, p := range inst.portals {
-		if p.id == id {
+	if int(id) >= 0 && int(id) < len(inst.portals) {
+		p := inst.portals[id]
+		if p.name != "" || p.destFieldID != constant.InvalidMap {
```
Hucaru: Do we need to check name or is invalid map all that's needed?

channel/field.go
```diff
@@ -785,15 +892,67 @@ func (inst fieldInstance) getPortalFromName(name string) (portal, error) {
 }
 
 func (inst fieldInstance) getPortalFromID(id byte) (portal, error) {
-	for _, p := range inst.portals {
-		if p.id == id {
+	if int(id) >= 0 && int(id) < len(inst.portals) {
```
Hucaru: MINOR: This if check is redundant as the index type is a byte and since this is an array of 256 length we can never be out of bounds.

channel/field.go
```diff
@@ -785,15 +892,67 @@ func (inst fieldInstance) getPortalFromName(name string) (portal, error) {
...
 			return p, nil
 		}
 	}
-
 	return portal{}, fmt.Errorf("No portal with that Name")
 }
 
+func (inst *fieldInstance) getNextPortalID() byte {
+	for i := 0; i < len(inst.portals); i++ {
+		if inst.portals[i].name == "" && inst.portals[i].destFieldID == constant.InvalidMap {
```
Hucaru, line 903: Do we need to check name as well as invalid map?

channel/field.go
```diff
@@ -785,15 +892,67 @@ func (inst fieldInstance) getPortalFromName(name string) (portal, error) {
...
+	}
+
+	return -1, &portal{}, fmt.Errorf("No available town portals")
+}
+
+// removePortalAtIndex removes a portal at the specified index
+func (inst *fieldInstance) removePortalAtIndex(index int) {
+	if index >= 0 && index < len(inst.portals) {
+		inst.portals[index] = portal{
+			id:          byte(index),
```
Hucaru: Do we need to create a new portal object? Can we not just set the invalid map vale to dest field to denote it as available for id re-use?

Issue #193, opened 2025-11-10
Implement HyperBody


ErwinsExpertise:
Currently, if you use Hyper Body it will show the increased HP locally, but it won't actually increase server side.

PR #202, opened 2025-11-17
Add Poison Mist, Mesos Explosion, and HyperBody

ErwinsExpertise:
#189 #192 #193 

This pull request introduces significant improvements to the buff, summon, and attack handling systems, as well as new infrastructure for managing map mist effects and supporting the Meso Explosion skill. The changes refactor buff expiration and restoration logic, streamline summon management (especially across map transfers), and add robust parsing and handling for complex attack types.

**Buff and Summon Management Improvements:**

- Refactored buff expiration logic to use a unified switch statement for skill-specific cleanup, including proper handling of `Recovery`, `HyperBody`, and all summon skills. Summon removal is now consistently managed through `removeActiveSummonForSkill`, and HP/MP adjustments for `HyperBody` are handled on buff expiration. [[1]](diffhunk://#diff-a6454022993b0895c14ed89b020f546bf952f4429fea2971610704be8cd2e230L1128-L1133) [[2]](diffhunk://#diff-a6454022993b0895c14ed89b020f546bf952f4429fea2971610704be8cd2e230L1157-L1179)
- Removed duplicate or obsolete summon-spawning logic from `AddBuff` and centralized it in `AddBuffFromCC`, ensuring that summons are only spawned when appropriate and with correct state. [[1]](diffhunk://#diff-a6454022993b0895c14ed89b020f546bf952f4429fea2971610704be8cd2e230L263-L287) [[2]](diffhunk://#diff-a6454022993b0895c14ed89b020f546bf952f4429fea2971610704be8cd2e230L856-R839)

**Map and Field Infrastructure:**

- Added a new `mistPool` to `fieldInstance` and integrated it into the field's lifecycle, enabling support for skills that create mist effects (e.g., Poison Mist). Players now see active mists when entering a map, and mists are updated with the field timer. [[1]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR554) [[2]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR340) [[3]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR677) [[4]](diffhunk://#diff-06e873df54ceb4e9fea4e908e64817448d5732c0ef07a7fb7df50861addb618aR975)

**Attack Parsing and Skill Handling:**

- Enhanced attack packet parsing to fully support the Meso Explosion skill, including per-target hit counts, mapping of meso drop object IDs to targets, and correct removal of mesos from the field. Added `handleMesoExplosion` to process the attack and apply damage only when mesos are consumed. [[1]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR2518-R2528) [[2]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL2477-R2675) [[3]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL2444-R2504) [[4]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR2358-R2365)
- Added support for parsing and creating mists for the Poison Mist skill in magic attack handling.

**Summon Persistence Across Map Transfers:**

- Improved handling of summons when a player changes maps: summons are now only kept if the corresponding buff is still active, and their state is restored or cleaned up as appropriate. [[1]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR1147-L1164) [[2]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL1173-R1182) [[3]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL891-R884)

**Other Notable Refactors:**

- Removed unused or obsolete functions such as `ClearBuff`, `buildBuffMask`, and `despawnSummonIfMatches` to simplify the codebase. [[1]](diffhunk://#diff-a6454022993b0895c14ed89b020f546bf952f4429fea2971610704be8cd2e230L991-L1004) [[2]](diffhunk://#diff-a6454022993b0895c14ed89b020f546bf952f4429fea2971610704be8cd2e230L1075-R1021) [[3]](diffhunk://#diff-a6454022993b0895c14ed89b020f546bf952f4429fea2971610704be8cd2e230L1157-L1179)
- Minor: Cleaned up player connection logic by removing forced return map checks that are now redundant.

These changes collectively improve the reliability, maintainability, and feature completeness of the server's buff, summon, and attack systems.

### DIFF
--- a/channel/field.go
+++ b/channel/field.go
@@ -502,6 +502,7 @@ func (f *field) changePlayerInstance(player *Player, id int) error {
 		}
 
 		f.instances[player.inst.id].dropPool.HideDrops(player)
+		f.instances[player.inst.id].mistPool.removeMistsByOwner(player.ID)
 
 		player.inst = f.instances[id]
 		err = f.instances[id].addPlayer(player)
--- a/channel/mist.go
+++ b/channel/mist.go
@@ -180,3 +180,14 @@ func (pool *mistPool) update(t time.Time) {
 		}
 	}
 }
+
+func (pool *mistPool) removeMistsByOwner(ownerID int32) {
+	if pool == nil {
+		return
+	}
+	for id, mist := range pool.mists {
+		if mist != nil && mist.ownerID == ownerID {
+			pool.removeMist(id)
+		}
+	}
+}


channel/field.go
Hucaru, line 1: Missing remove mist when player switches instances on the same field?

channel/handlers_client.go
```diff
@@ -2441,7 +2454,54 @@ func (server Server) playerMagicSkill(conn mnet.Client, reader mpacket.Reader) {
...
+	for tIdx, at := range data.attackInfo {
+		found := false
+		for _, did := range at.mesoDropIDs {
+			if _, ok := existing[did]; ok {
+				found = true
+				break
+			}
+		}
+		hadMappedDrop[tIdx] = found
+	}
```
Hucaru: It feels like there should be a way to combine the above 3 loops

channel/handlers_client.go
```diff
@@ -2474,88 +2537,142 @@ func getAttackInfo(reader mpacket.Reader, player Player, attackType int) (attack
...
 		data.option = reader.ReadByte()
 
 		tmp := reader.ReadByte()
 		data.action = tmp & 0x7F
 		data.facesLeft = (tmp >> 7) == 1
 		data.attackType = reader.ReadByte()
 
-		reader.Skip(4) // checksum
+		// The client sends a 4-byte field here (checksum/random). Keep alignment by consuming it.
+		_ = reader.ReadInt32()
```
Hucaru: Why change from skip to read?

channel/mist.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func createNewMistPool(inst *fieldInstance) mistPool {
+	return mistPool{
+		instance: inst,
+		mists:    make(map[int32]*fieldMist),
+	}
+}
+
+func (pool *mistPool) nextID() int32 {
```
Hucaru, line 46: I think this can just be a simple increment. To be honest the change that max int32 and 0 are both going to be active is nearly impossible due to how fast mist times out and number of people on field along with how much they can spam the skill

channel/monster.go
```diff
@@ -299,23 +300,46 @@ func (m monster) String() string {
...
 	if m.hp <= 0 {
 		return
 	}
 
-	if m.poison {
-		// Handle poison (TODO: scale by poison level)
-		m.hp -= 10
-		if m.hp < 0 {
-			m.hp = 0
+	if (m.statBuff&skill.MobStat.Poison) != 0 && m.buffs != nil {
```
Hucaru, line 311: Mist cannot kill, if mob is at hp < 2 then it cannot take mist damage.

channel/field.go
ErwinsExpertise: Hmm, so not a map switch, but a completely different instance? That's not something I really thought about since it's not consistent across the source. Is there specific logic you had in mind? I can't think of any real reasons to clear it out from the instance though. 

Maybe?
```
		if plr, _ := pool.instance.getPlayerFromID(mist.ownerID); plr == nil {
			pool.removeMist(id)
		}
```

Hucaru: So it's up to you where you want to do it as the game was never built to allow switching instances of the same map whilst being in it:

Call it directly here (where I think it's most appropriate): https://github.com/Hucaru/Valhalla/blob/master/channel/field.go#L491C17-L491C37

Or in https://github.com/Hucaru/Valhalla/blob/master/channel/field.go#L777 and adding a check for portal use being false.

> That's not something I really thought about since it's not consistent across the source

I think this is the first case where something persists in a field? In other situations e.g. boats I just iterate over the instances and send the packets.

channel/monster.go
```diff
@@ -335,6 +343,10 @@ func (m *monster) update(inst *fieldInstance, t time.Time) {
 					}
 
 					inst.lifePool.mobDamaged(m.spawnID, nil, dmg)
+
+					if m.hp <= 1 {
```
Hucaru, line 347: Can this be reached due to line 321?

ErwinsExpertise: Yes, this is a check _after_ the damage occurs, while the previous one is a check before damage.

channel/field.go
ErwinsExpertise: Added mist cleanup in `changePlayerInstance` as suggested.

PR #206, opened 2025-11-27
Adjust readme for metrics

ErwinsExpertise:
This pull request updates the installation and local setup documentation to clarify instructions and add important configuration details, especially regarding schema import locations and metrics port configuration for running multiple server instances.

Documentation improvements:

* Updated command examples in `docs/Installation.md` to use the correct binary names (`wztonx-converter.exe`/`wztonx-converter`) and flags, and clarified usage instructions.
* Clarified the directory for importing the SQL schema in `docs/Installation.md`, specifying the "Valhalla SQL directory" for better accuracy.
* Improved instructions in `docs/Local.md` by specifying that the schema should be imported from the SQL directory.

Metrics port configuration:

* Added a new section to `docs/Local.md` explaining the need to specify unique `-metrics-port` values when running multiple channels or the cash shop server locally, with updated example commands for each server instance.

### DIFF
--- a/docs/Local.md
+++ b/docs/Local.md
@@ -185,7 +185,7 @@ Valhalla.exe -type channel -config config_channel_1.toml
 ```
 
 #### Additional Channels/CashShop 
-By default Prometheus uses port 9000 for metrics. When running locally it is not able to bind to the same port so you will need to specify alternative metrics ports when running then binary `-metrics-port 9xxx`. It's easiest to increment 1 for every channel: 9001, 9002, 9003, etc. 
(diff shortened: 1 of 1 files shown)


docs/Local.md
```diff
@@ -184,22 +184,24 @@ Valhalla.exe -type channel -config config_channel_1.toml
 ./Valhalla -type channel -config config_channel_1.toml
 ```
 
+#### Additional Channels/CashShop 
+By default Prometheus uses port 9000 for metrics. When running locally it is not able to bind to the same port so you will need to specify alternative metrics ports when running then binary `-metrics-port 9xxx`. It's easiest to increment 1 for every channel: 9001, 9002, 9003, etc.
```
Hucaru: when running then binary -> when running the binary

PR #215, opened 2025-12-07
Feature/event system

### LINKED ISSUE #105: Kerning PQ

### LINKED ISSUE #219: Party HP not sync'd on field change
Party HP is not sync'd on field transitions

Hucaru:
Added event system with kerning pq as an example. I expect there will be edge cases I have not caught. They should go in as new PRs. Closes #105 closes #219

Work still to be done:
- Flavour text for npcs. Some has already done by looking at videos and using https://www.imagetotext.info/ to get dialogue text.
- Correct reward amounts for potions/elixirs/ores

514145149-3538a2a6-6d30-471a-abf6-2df822db6487

### DIFF
--- a/cashshop/handle_client.go
+++ b/cashshop/handle_client.go
@@ -219,6 +219,105 @@ func (server *Server) playerCashShopPurchase(conn mnet.Client, reader mpacket.Re
 
 		plr.Send(packetCashShopUpdateAmounts(plrNX, plrMaplePoints))
 
+	case opcode.RecvCashShopBuyPackage, opcode.RecvCashShopGiftPackage:
+		currencySel := reader.ReadByte()
+		pkgSN := reader.ReadInt32()
+
+		commodity, ok := nx.GetCommodity(pkgSN)
+		if !ok || commodity.Price <= 0 {
+			plr.Send(packetCashShopUpdateAmounts(plrNX, plrMaplePoints))
+			return
+		}
+
+		pkgMap := nx.GetPackages()
+		pkgItems, ok := pkgMap[pkgSN]
+		if !ok || len(pkgItems) == 0 {
+			// Fallbacks: some data sets key packages by ItemID or Commodity index instead of SN
+			if commodity.ItemID != 0 {
+				pkgItems, ok = pkgMap[commodity.ItemID]
+			}
+			if (!ok || len(pkgItems) == 0) && commodity.Index != 0 {
+				pkgItems, ok = pkgMap[commodity.Index]
+			}
+		}
+		if !ok || len(pkgItems) == 0 {
+			plr.Send(packetCashShopUpdateAmounts(plrNX, plrMaplePoints))
+			return
+		}
+
+		price := commodity.Price
+		switch currencySel {
+		case constant.CashShopNX:
+			if plrNX < price {
+				plr.Send(packetCashShopError(opcode.SendCashShopBuyFailed, constant.CashShopErrorNotEnoughCash))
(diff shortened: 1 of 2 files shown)


channel/script.go
```diff
@@ -598,9 +737,17 @@ func (ctrl *npcChatController) Send(text string) int {
 }
 
 // SendBackNext packet to Player
-func (ctrl *npcChatController) SendBackNext(msg string, back, next bool) {
```
Hucaru, line 601: @ErwinsExpertise This will invalidate a lot of NPCs. Once the current changes are approved I will edit the other scripts and push them up to prevent the PR diff from getting out of hand.

scripts/npc/9020001.js
```diff
@@ -1,152 +1,218 @@
...
-            plr.instanceProperty("stage1a", 1);
-        } else if (plr.instanceProperty("stage1a") === "1") {
+            props.stage = 1;
+        } else if (props.stage === 1) {
             npc.sendNext("I'm sorry, but at least one party member still hasn't completed their mission. Everyone except the part leader must clear their mission to move on.");
-        } else if (plr.instanceProperty("stage1a") === "2") {
+        } else if (props.stage === 2) {
+            props.stage = 3;
+            clear();
+            plr.partyGiveExp(100);
```
Hucaru, line 117: I think stage 1 might have a formula applied either to the party or per player based on the number of coupons collected?

channel/event.go
```diff
@@ -0,0 +1,183 @@
...
+		vm:         goja.New(),
+	}
+
+	ctrl.vm.SetFieldNameMapper(goja.UncapFieldNameMapper())
+	_ = ctrl.vm.Set("ctrl", ctrl)
+
+	_, err := ctrl.vm.RunProgram(ctrl.program)
+
+	if err != nil {
+		log.Println(err)
```
ErwinsExpertise: Do we want to just log and continue after failing to run the program? It makes sense to me to return an error instead and stop initialization.

channel/event.go
```diff
@@ -0,0 +1,183 @@
...
+func (e *event) RemainingTime() int32 {
+	return int32(time.Until(e.endTime).Seconds())
+}
+
+func (e *event) PlayerCount() int {
+	return len(e.playerIDs)
+}
+
+func (e *event) Finished() {
+	close(e.finished)
```
ErwinsExpertise: Should probably use sync.Once here to prevent possible panics.

channel/event.go
```diff
@@ -0,0 +1,183 @@
...
+	err = ctrl.vm.ExportTo(ctrl.vm.Get("timeout"), &ctrl.timeoutCallback)
+
+	if err != nil {
+		log.Println(err)
+	}
+
+	return ctrl
+}
+
+func (e *event) start(server *Server) {
```
ErwinsExpertise: The `server` param is unnecessary since `event` struct already contains pointer `e.server`

channel/event.go
```diff
@@ -0,0 +1,183 @@
...
+	e.startCallback()
```
ErwinsExpertise, line 91: Should this be nil checked?

Hucaru: Not needed, there is an error check at event creation for this function being defined.

channel/event.go
Hucaru: Yes, this should return an error

channel/event.go
Hucaru: Good catch

PR #216, opened 2025-12-10
fix recovery skill buff

luizxd0:
Fixed the problem with HP Sync when using beginner skill Recovery


ErwinsExpertise:
@Hucaru this seems to pop up way more often than it should. Might need to refactor buffs so that it doesn’t use a player pointer. Admittedly, the current implementation is a bit clunkier than I would have liked, but I think just resetting the pointer every time is an easy patch until it can get refactored.

PR #217, opened 2025-12-10
fix random items quest.

luizxd0:


channel/player.go
```diff
@@ -2165,7 +2159,7 @@ func (d *Player) applyQuestAct(act nx.ActBlock, npcID int32, questID int16) erro
 		}
 	}
 
-	if hasRandomReward {
+	if hasRandomReward && totalWeight > 0 {
```
Hucaru:
```
hasRandomReward = true
totalWeight += ai.Prop
```
Can `hasRandomReward` be true if `totalWeight == 0`? If not then you only need `totalWeight`

luizxd0: I honestly dont know. I'm not a real dev. I use Cursor with claude or gpt5.1 to code. That's how it fixed the issue, I tested and now generates a random item correctly

Hucaru: Try removing line 2138 and change line 2162 to `if totalWeight > 0 {` and test those changes

luizxd0: I tried that and it keeps giving the same item over and over.
The fix I applied works fine tho
image

luizxd0: Actually after testing my own code, it started giving the same item over and over. 

The quest reward roll was using the per-player RNG, which is deterministically seeded from stable IDs. That means every time the server restarted (or for players who hadn’t advanced their RNG state), the first call to that RNG produced the same number, so the weighted prize picker kept landing on the same item. I switched the roll to a fresh time-seeded math/rand instance per reward selection, so each quest turn-in now gets a different seed and produces varied picks.

Here is the change:

image

I also followed your suggestion and removed the hasrandomreward and it works fine now. Looks like this is what fixed the issue:

		rng := rand.New(rand.NewSource(time.Now().UnixNano()))
		roll := int32(rng.Int31n(totalWeight))

channel/player.go
```diff
@@ -2167,8 +2159,9 @@ func (d *Player) applyQuestAct(act nx.ActBlock, npcID int32, questID int16) erro
 		}
 	}
 
-	if hasRandomReward {
-		roll := int32(d.randIntn(int(totalWeight)))
+	if totalWeight > 0 {
+		rng := rand.New(rand.NewSource(time.Now().UnixNano()))
```
Hucaru: @ErwinsExpertise are you ok with this bypassing the deterministic seed?

ErwinsExpertise: This is fine to me, but it's not exactly all that random if 2 players are using the same seed since it's based on time. Maybe switch to crypto/rand? 

```
var seed int64
binary.Read(crypto_rand.Reader, binary.BigEndian, &seed)
rng := rand.New(rand. NewSource(seed))
```

luizxd0: done, is it better now?

ErwinsExpertise (approved):
lgtm

PR #220, opened 2025-12-17
fixed npc send next, back, next & back function calls

Hucaru:
Event system changed function name and params for npc send next, back and back & next dialogue boxes


Hucaru:
@ErwinsExpertise this is good for review

PR #234, opened 2026-01-01
Add player shops

ErwinsExpertise:
This pull request implements the core logic and packet handling for player-owned shops (personal shops) in the game, allowing players to create, manage, and interact with shop rooms. It also introduces a more flexible interface for displaying room/game boxes and makes several related improvements and bug fixes to inventory and shop item handling.

**Personal Shop System Implementation:**

* Added handling for creating, opening, closing, and removing personal shop rooms, including reading shop data from the client, managing shop state, and updating the room pool accordingly (`channel/handlers_client.go`, `channel/pools.go`). [[1]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL3251-R3268) [[2]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR3341-R3352) [[3]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR3379-R3397) [[4]](diffhunk://#diff-c895621c7ad98ab540b512ea81976404587f14687fc7b492bde56f2ca594c99eR884-R891)
* Implemented logic for adding items to a shop, buying items from a shop, and moving items from the shop back to the owner's inventory, complete with appropriate checks and packet responses (`channel/handlers_client.go`).
* Defined new constants for shop-related packet operations and increased the maximum number of players in a shop (`constant/constants.go`). [[1]](diffhunk://#diff-29ca518091dc3acef302068df3fe17daab605542298d7be4d840b61bdae9433aR333) [[2]](diffhunk://#diff-29ca518091dc3acef302068df3fe17daab605542298d7be4d840b61bdae9433aR371-R375)

**Inventory and Item Handling Improvements:**

* Refactored `updateItemStack` to accept a `silent` parameter, preventing unnecessary packet sends when not needed, and updated all usages accordingly (`channel/player.go`, `channel/handlers_client.go`). [[1]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL3153-R3153) [[2]](diffhunk://#diff-7c17860a8d1d10607463092eb6652bafd66b1f34027f14eeb42fa052a1385971L1182-R1189) [[3]](diffhunk://#diff-7c17860a8d1d10607463092eb6652bafd66b1f34027f14eeb42fa052a1385971L1210-R1227)
* Renamed and adjusted logic for taking items silently (for storage/shop operations), ensuring correct removal and update of rechargeable and non-rechargeable items (`cashshop/handle_client.go`, `channel/player.go`). [[1]](diffhunk://#diff-f6913926b5d774f61e1b9ae690b91becce7929faa0eaaf1997951ae4e9483e87L476-R476) [[2]](diffhunk://#diff-7c17860a8d1d10607463092eb6652bafd66b1f34027f14eeb42fa052a1385971L1182-R1189) [[3]](diffhunk://#diff-7c17860a8d1d10607463092eb6652bafd66b1f34027f14eeb42fa052a1385971L1210-R1227)
* Fixed handling of rechargeable items when moving them into trade/shops to ensure proper removal from inventory (`channel/handlers_client.go`).

**Room/Game Box Display Refactor:**

* Unified the interface for displaying room/game boxes by introducing and using a `boxDisplayer` interface instead of multiple type assertions (`channel/pools.go`). [[1]](diffhunk://#diff-c895621c7ad98ab540b512ea81976404587f14687fc7b492bde56f2ca594c99eL790-R792) [[2]](diffhunk://#diff-c895621c7ad98ab540b512ea81976404587f14687fc7b492bde56f2ca594c99eL837-R838) [[3]](diffhunk://#diff-c895621c7ad98ab540b512ea81976404587f14687fc7b492bde56f2ca594c99eL865-R868)

**Minor Fixes:**

* Corrected the displayed channel number when registering a new channel (`channel/handlers_server.go`).
* Cleaned up SQL schema file formatting (`sql/maplestory.sql`).


channel/handlers_client.go
```diff
@@ -3601,6 +3641,131 @@ func (server Server) roomWindow(conn mnet.Client, reader mpacket.Reader) {
 				}
 			}
 		}
+	case constant.MiniRoomAddShopItem:
```
ErwinsExpertise, line 3660: Test case I need to try - adding the same item twice. This PR does not “take” items to put them up for sale that way we don’t have to worry about rollbacks, but are players able to add the same item twice? If so logic will need to be modified to prevent the behavior

ErwinsExpertise: The answer is yes, and this is super buggy. So I just changed it to use escrow instead.

ErwinsExpertise:
> Why are player shops being stored in the database? These only need to persist whilst a player is online. Only hired merchants need to persist.
> 
> The player struct could be modified to have a `shopItems []*Item` (fixed size array prefered over slice if possible) entry where `salePrice` and `saleAmount` are added to the item struct. This way if a player disconnects there is nothing to manage apart from removing the room from the pool.
> 
> I can't remember how quantites are managed in player shops but if the client supports selling more than 1 of an item then at the point of sale the `saleAmount` and `amount` fields in the item can both be modified and the item can then be marked as dirty/saved.

I went with DB escrow because once an item is listed I’m actually removing it from the character’s inventory, and keeping that removed inventory purely in memory means a crash can lose items. I originally tried the simpler “visual remove / add back later” approach and ran into consistency problems: the client inventory UI and the server’s idea of slots/stacks would drift, and getting items to reliably reappear without desyncs or edge-case glitches was messy. Escrow keeps things authoritative and predictable by making the item live in exactly one place at a time (inventory or escrow) and gives a clean restore path on login. It also sets things up for hired merchants later since they need persistence and can reuse the same escrow/restore mechanism instead of introducing a separate system.

Hucaru:
> I went with DB escrow because once an item is listed I’m actually removing it from the character’s inventory, and keeping that removed inventory purely in memory means a crash can lose items. I originally tried the simpler “visual remove / add back later” approach and ran into consistency problems: the client inventory UI and the server’s idea of slots/stacks would drift, and getting items to reliably reappear without desyncs or edge-case glitches was messy. Escrow keeps things authoritative and predictable by making the item live in exactly one place at a time (inventory or escrow) and gives a clean restore path on login. It also sets things up for hired merchants later since they need persistence and can reuse the same escrow/restore mechanism instead of introducing a separate system.

This should only be the case if a save isn't performed at the point of sale. When an item is bought a save should occur for both parties, to guarantee no desync's. If the seller disconnects and relogs then they will just load in the inventory from db which will be pre-shop creation minus any sales.

When a player makes a shop and places an item into the shop list what happens in the client UI? Does the item disappear from it's item slot like a trade window? Are you able to attach a screenshot? You should be able to treat this similar to a trade room with only 1 side having items.

ErwinsExpertise:
> > I went with DB escrow because once an item is listed I’m actually removing it from the character’s inventory, and keeping that removed inventory purely in memory means a crash can lose items. I originally tried the simpler “visual remove / add back later” approach and ran into consistency problems: the client inventory UI and the server’s idea of slots/stacks would drift, and getting items to reliably reappear without desyncs or edge-case glitches was messy. Escrow keeps things authoritative and predictable by making the item live in exactly one place at a time (inventory or escrow) and gives a clean restore path on login. It also sets things up for hired merchants later since they need persistence and can reuse the same escrow/restore mechanism instead of introducing a separate system.
> 
> This should only be the case if a save isn't performed at the point of sale. When an item is bought a save should occur for both parties, to guarantee no desync's. If the seller disconnects and relogs then they will just load in the inventory from db which will be pre-shop creation minus any sales.
> 
> When a player makes a shop and places an item into the shop list what happens in the client UI? Does the item disappear from it's item slot like a trade window? Are you able to attach a screenshot? You should be able to treat this similar to a trade room with only 1 side having items.

I get the just save on sale argument, but that doesn’t address what was actually going wrong for me. The hard part wasn’t persisting the buyer/seller at the moment of purchase, it was trying to model a shop as “the item is still in the owner’s inventory, but also listed somewhere else” and then relying on UI tricks or a trade-like one-sided flow to keep everything consistent. Once you do that you’re forced into either faking inventory state (and dealing with slot/stack desyncs when things merge/split, partial quantities, etc.) or implementing a real lock/reserve system so the listed item can’t be moved/dropped/traded/consumed while it’s listed. That lock approach sounds simple but it spreads everywhere: every inventory-changing path has to check and enforce it, and if you miss one you’ve created a dupe/desync. The DB escrow approach keeps it contained and deterministic: listing is a real server-side state change (item is no longer in inventory while listed), so you don’t need global locking rules, restore is straightforward, and it’s also the same primitive hired merchants will need anyways. So this isn’t persisting player shops, it’s persisting removed inventory safely and reusing the same path for merchant persistence later.

ErwinsExpertise:
Okay @Hucaru 

I implemented virtual stack-splitting: when listing a partial amount from a stack, we don’t actually split or reduce the owner’s inventory item up-front. Instead each shop listing tracks a `reserved` amount (keyed by the item’s `dbID`), and we send inventory-operation packets to visually show the remaining count (or hide the stack if fully reserved). On purchase, we validate reservations vs the owner’s current real stack, then `TakeItemSilent` the purchased amount and decrement the listing’s `reserved/bundles`. After any buy that changes listings, we normalize shop slots to stay `0..N-1` so the client’s index-based slot selection remains consistent.

channel/handlers_client.go
```diff
@@ -3347,7 +3376,25 @@ func (server Server) roomWindow(conn mnet.Client, reader mpacket.Reader) {
 			if err != nil {
 				log.Println(err)
 			}
+		} else if shop, valid := r.(*shopRoom); valid {
```
Hucaru: I think this if else if would now be better as a switch on interface type

channel/handlers_client.go
```diff
@@ -3362,12 +3409,13 @@ func (server Server) roomWindow(conn mnet.Client, reader mpacket.Reader) {
 
 		if item.isRechargeable() {
 			amount = item.amount
-		}
-
-		_, err = plr.takeItem(item.ID, invSlot, amount, invType)
-		if err != nil {
-			plr.Send(packetPlayerNoChange())
-			return
+			plr.removeItem(item, false)
```
Hucaru: Should we not remove the item after checking there is space to insert it?

channel/handlers_client.go
```diff
@@ -3601,6 +3649,92 @@ func (server Server) roomWindow(conn mnet.Client, reader mpacket.Reader) {
...
+		if err != nil {
+			plr.Send(packetPlayerNoChange())
+			return
+		}
+
+		if shop.addItem(item, bundles, bundleAmount, price, shop.nextSlot) {
+			shop.send(packetRoomShopRefresh(shop))
+		}
+
+		plr.Send(packetPlayerNoChange())
```
Hucaru, line 3686: Do we need to send this? Can a legit player fail the above if check?

channel/handlers_client.go
```diff
@@ -3601,6 +3649,92 @@ func (server Server) roomWindow(conn mnet.Client, reader mpacket.Reader) {
...
+		if !valid {
+			return
+		}
+
+		if shop.ownerID() == plr.ID {
+			return
+		}
+
+		if !shop.checkOpen() {
+			// This shouldn't be necessary, but we do it anyways just in-case :)
```
Hucaru, line 3707: Under what situations is this required?

channel/room.go
```diff
@@ -926,6 +939,469 @@ func (r *tradeRoom) rollback() {
...
+func (r *shopRoom) ownerPlayer() *Player {
+	for _, plr := range r.players {
+		if plr != nil && plr.ID == r.ownerPlayerID {
+			return plr
+		}
+	}
+	return nil
+}
+
+func (r *shopRoom) closeShop(reason byte, owner *Player) {
```
Hucaru: You don't need to pass in the owner player?

channel/room.go
```diff
@@ -926,6 +939,469 @@ func (r *tradeRoom) rollback() {
...
+			slotID int16
+		}
+		seen := make(map[key]struct{}, len(r.items))
+
+		for _, si := range r.items {
+			if si == nil || si.item.dbID == 0 {
+				continue
+			}
+			k := key{dbID: si.item.dbID, invID: si.item.invID, slotID: si.item.slotID}
+			if _, ok := seen[k]; ok {
```
Hucaru: Is protecting from telling the client it has more items when a stack was split multiple times and they need to rejoin? I think having the key just be the DB id would be enough?

channel/room.go
```diff
@@ -926,6 +939,469 @@ func (r *tradeRoom) rollback() {
...
+			}
+		}
+	}
+
+	r.items = make(map[byte]*shopItem)
+	r.nextSlot = 0
+	r.players = []*Player{}
+}
+
+func (r *shopRoom) checkOpen() bool {
```
Hucaru: I'm not sure what purpose this serves. Player disconnects or player manually closes shops are already handled?

channel/room.go
```diff
@@ -926,6 +939,469 @@ func (r *tradeRoom) rollback() {
...
+	reserved     int16
+	hidden       bool
+}
+
+type shopRoom struct {
+	room
+	name     string
+	private  bool
+	open     bool
+	items    map[byte]*shopItem
```
Hucaru: I'm assuming there is a max item count? If so then a slice seems more appropriate and memory efficient

channel/room.go
```diff
@@ -926,6 +939,469 @@ func (r *tradeRoom) rollback() {
...
+	if !buyer.canReceiveItems([]Item{purchased}) {
+		return constant.PlayerShopInventoryFull, false
+	}
+
+	buyer.takeMesos(int32(totalCost))
+	owner.giveMesos(int32(totalCost))
+
+	if err, _ := buyer.GiveItem(purchased); err != nil {
+		owner.takeMesos(int32(totalCost))
+		buyer.giveMesos(int32(totalCost))
```
Hucaru: owner receive and buyer take?

channel/room.go
```diff
@@ -926,6 +939,469 @@ func (r *tradeRoom) rollback() {
...
+	}
+
+	for _, si := range r.items {
+		if si != nil && si.item.dbID == dbID {
+			si.hidden = false
+		}
+	}
+}
+
+func (r *shopRoom) normalizeShopSlots() {
```
Hucaru: Is this accounting for something in the ui? Do the items move up a slot if one slot gets emptied? If so then this would be better served as a slice as this is trivial to do.

channel/handlers_client.go
ErwinsExpertise: Not really, but I think if it fails then the client will get stuck and that's not really a good state to leave it in even if it is a cheater.

channel/handlers_client.go
ErwinsExpertise: It's not really required, it's just to be 100% sure that a shop is open before committing to the sale. Most all situations should be accounted for, but I think due to the nature of shops it's not a bad thing to be safe.

channel/room.go
ErwinsExpertise: I can't find anything on the max, but it is a byte so nothing more than byte max technically. I think it's likely wayyy lower than that like 32, but I'm not sure.

channel/room.go
ErwinsExpertise: This is for a rollback

channel/room.go
```diff
@@ -926,6 +960,441 @@ func (r *tradeRoom) rollback() {
...
+
+		for _, si := range r.items {
+			if si == nil || si.item.dbID == 0 {
+				continue
+			}
+			id := si.item.dbID
+			if _, ok := seen[id]; ok {
+				continue
+			}
+			seen[id] = struct{}{}
```
Hucaru: I think there is a way of doing this without having to do a map lookup and insertion.

Can we not send the client each item in the shop as it should add up to what they had minus what was sold?

channel/room.go
```diff
@@ -926,6 +960,441 @@ func (r *tradeRoom) rollback() {
...
+	}
+
+	for _, si := range r.items {
+		if si != nil && si.item.dbID == dbID {
+			si.hidden = false
+		}
+	}
+}
+
+func (r *shopRoom) syncShopItemSlots() {
```
Hucaru: You can use the slice index as the slot. When an item is sold and the slots need to be moved up you can do:
```
slices.Delete(r.items, slot, slot+1)
```
Example at https://go.dev/play/p/8G-RHlWjHqr

channel/room.go
Hucaru, line 1: I wonder if this file should be split into the following files:
room.go -> general room struct, interfaces and packets
room_shop.go -> shop related structs, function and packets
room_game.go -> ...
room_trade.go -> ...

thoughts @ErwinsExpertise ?

ErwinsExpertise: Great idea! I went ahead and moved them into their own files.

channel/room_shop.go
```diff
@@ -0,0 +1,459 @@
...
+			if plr.Conn == nil {
+				r.closeShop(constant.MiniRoomClosed)
+				return false
+			}
+		}
+	}
+	return true
+}
+
+func (r *shopRoom) reservedForDBID(dbID int64) int16 {
```
Hucaru, line 132: I'm not sure I follow what the intent of this function is. Can you explain it's purpose (It has something to do with items that can be stacked)?

ErwinsExpertise: I added `reservedForDBID` to properly handle stackable items in the shop. A single inventory stack (same `dbID`) can be listed across one or more shop entries, so we keep a “reserved” count per entry and then sum it with `reservedForDBID` to know how much of that underlying stack is already committed to the shop at any moment. That lets us (1) prevent over-listing / overselling the same stack (including desync/hack attempts where the client tries to buy more than should exist), and (2) keep the owner’s inventory view consistent by calculating “available = current amount - reserved” and hiding the stack when everything is reserved, then restoring it when reservations drop.

Hucaru: Thanks I assumed it was something to do with stack verification but I wanted to know the cases.

PR #240, opened 2026-01-16
small fixes for LPQ

ErwinsExpertise:
This pull request updates the Ludibrium Party Quest (LudiPQ) event flow and related script logic, mainly to improve the handling of bonus and exit maps, clarify player transitions, and refactor party warping functionality. The changes ensure players are correctly routed between bonus, exit, and starting maps, and improve the clarity and flexibility of related NPC dialogues and event timeouts.

**Event and Party Flow Improvements:**

* Changed the exit map for the main event to `922010000` and introduced a separate `bonusExitMapID` (`922011100`) for the bonus stage, ensuring players are sent to the correct map after the event or bonus stage.
* Updated the event timeout logic to warp players to the bonus exit map instead of the regular exit, aligning with the new flow.

**Party Warping and Experience Logic:**

* Modified `PartyWarp` to accept both source and destination map IDs, and updated its usage to warp only party members in the correct map (prevents unintended warps). [[1]](diffhunk://#diff-432dd333cd400bc1a1e8651dc65ad5f71be0b2cab06ec7e1205d1b49f7b6cd99L353-R353) [[2]](diffhunk://#diff-432dd333cd400bc1a1e8651dc65ad5f71be0b2cab06ec7e1205d1b49f7b6cd99R374-R379) [[3]](diffhunk://#diff-543ace4b4ef6440fa6d82252026b9c210db0b614a2f13ebbabe0a2e3a0cb962dL15-R15)

**NPC and Player Dialogue Enhancements:**

* Enhanced NPC 2040045's dialogue to clarify the bonus stage and provide a more thematic exit prompt, while ensuring players are warped to the correct map upon leaving.
* Improved NPC 2040047's exit prompt to distinguish between leaving from the exit room versus other maps, and adjusted item cleanup and warp destinations accordingly.

**Event Timer Robustness:**

* Refactored the event timer logic in `channel/event.go` to use a loop and ensure proper cleanup and return behavior, improving reliability when handling timeouts and timer resets. [[1]](diffhunk://#diff-b49b8e8557bb7a0dd87cd2796eaf715b4fb8618445f197cf074c58416f8078eaR116) [[2]](diffhunk://#diff-b49b8e8557bb7a0dd87cd2796eaf715b4fb8618445f197cf074c58416f8078eaR129-R130) [[3]](diffhunk://#diff-b49b8e8557bb7a0dd87cd2796eaf715b4fb8618445f197cf074c58416f8078eaR152)

**Minor Cleanup:**

* Removed an unnecessary blank line in the bonus exit NPC script for consistency.

### DIFF
--- a/channel/script.go
+++ b/channel/script.go
@@ -338,6 +338,20 @@ func (ctrl *scriptPlayerWrapper) PartyMembersOnMap() []scriptPlayerWrapper {
 	return members
 }
 
+func (ctrl *scriptPlayerWrapper) EventMembersOnMap(id int32) bool {
+	if ctrl.plr.event == nil {
+		return false
+	}
+
+	return ctrl.plr.event.IsParticipantsOnMap(id)
+}
+
+func (ctrl *scriptPlayerWrapper) WarpEventMembers(id int32) {
+	if ctrl.plr.event != nil {
+		ctrl.plr.event.WarpPlayers(id)
+	}
+}
+
 func (ctrl *scriptPlayerWrapper) PartyGiveExp(val int32) {
 	if !ctrl.InParty() {
 		return
--- a/scripts/npc/2040044.js
+++ b/scripts/npc/2040044.js
@@ -5,7 +5,7 @@ var props = map.properties();
 if (!plr.isPartyLeader()) {
     npc.sendOk("Here is the information about the 9th stage. Now is your chance to finally get your hands on the real culprit. Go right and you'll see a monster. Defeat it to find a monstrous #b#o9300012##k appearing out of nowhere. Your task is to defeat him, collect the #b#t4001023##k he has and bring it to me.");
 } else if (plr.itemCount(key) >= 1) {
-    if (ctrl.isParticipantsOnMap(plr.mapID())) {
+    if (plr.eventMembersOnMap(plr.mapID())) {
         if (npc.sendYesNo("Incredible! You defeated Alishar and obtained the #t" + key + "#! Would you like to proceed to the bonus stage?")) {
             plr.removeItemsByID(key, 1);
             props.clear = true;
@@ -13,7 +13,7 @@ if (!plr.isPartyLeader()) {
             map.playSound("Party1/Clear");
             map.portalEffect("gate");
             plr.partyGiveExp(27200);
-            ctrl.warpPlayers(922011000);
+            plr.warpEventMembers(922011000);
         }
     } else {
         npc.sendOk("Please make sure all members of your party are in the current map.")


channel/script.go
```diff
@@ -350,7 +350,7 @@ func (ctrl *scriptPlayerWrapper) PartyGiveExp(val int32) {
 	}
 }
 
-func (ctrl *scriptPlayerWrapper) PartyWarp(dst int32) {
+func (ctrl *scriptPlayerWrapper) PartyWarp(src, dst int32) {
```
Hucaru: Only warp party players if they are on src map?

scripts/npc/2040044.js
```diff
@@ -12,7 +12,7 @@ if (!plr.isPartyLeader()) {
         map.playSound("Party1/Clear");
         map.portalEffect("gate");
         plr.partyGiveExp(27200);
-        plr.partyWarp(922011000);
+        plr.partyWarp(922010900, 922011000);
```
Hucaru: What happens if a player is in a previous stage? What is the intended behaviour for them?

channel/script.go
ErwinsExpertise: If a player left PQ they would be warped so this is to prevent that behavior.

scripts/npc/2040044.js
ErwinsExpertise: They wouldn't get warped. I think it might be a good idea to check that all players are on map prior to allowing clear. That's how it worked in GMS iirc, so it makes sense here too.

ErwinsExpertise: Is it preferable to just check for party on map before clear, or should we check src for warp too? I'm ok either way.

channel/script.go
```diff
@@ -307,6 +307,20 @@ func (ctrl *scriptPlayerWrapper) IsPartyLeader() bool {
 	return false
 }
 
+func (ctrl *scriptPlayerWrapper) IsPartyOnMap() bool {
```
Hucaru: This will block PQ progress if 1 player leaves which is allowed.

ErwinsExpertise: Do they get removed from event when they leave? We could check against event status too if that's the case.

ErwinsExpertise: @Hucaru I'm thinking:

```
	for _, v := range ctrl.plr.party.players {
		if v.mapID != ctrl.MapID() && v.event != nil {
			return false
		}
	}
```

and for warp

```
	for _, plr := range ctrl.plr.party.players {
		if plr != nil && plr.event != nil {
			ctrl.server.warpPlayer(plr, field, dstPortal, false)
		}
	}
```

channel/script.go
Hucaru: I would rename this to EventWarp() otherwise it can mixed up with warp a party even if not in event.

channel/script.go
Hucaru: I think you want this to be something like:
```
func (ctrl *scriptPlayerWrapper) EventParticipantsOnMap() {
 if !ctrl.plr.event {
   return false
 }

  // Iterate over event participants and if one of them is not on map return false
}

Comment (ErwinsExpertise) [reply channel/script.go]: Because it returns a bool I think it would be better to have the function prepended with `Is` to prevent confusion on what it's for. I added the owner event check as well to method and renamed them.

File: channel/script.go [hunk 4]
```diff
@@ -307,9 +307,9 @@ func (ctrl *scriptPlayerWrapper) IsPartyLeader() bool {
...
 
-func (ctrl *scriptPlayerWrapper) IsPartyOnMap() bool {
-	if !ctrl.InParty() {
-		return true
+func (ctrl *scriptPlayerWrapper) IsEventParticipantsOnMap() bool {
+	if ctrl.plr.event == nil {
+		return false
 	}
 
 	for _, v := range ctrl.plr.party.players {
```
Hucaru: Need to iterate over event.playerIDs to check if they are on same mapID. Can't be party based as this wouldn't work for GM events.

channel/script.go
```diff
@@ -364,7 +364,7 @@ func (ctrl *scriptPlayerWrapper) PartyGiveExp(val int32) {
 	}
 }
 
-func (ctrl *scriptPlayerWrapper) PartyWarp(dst int32) {
+func (ctrl *scriptPlayerWrapper) EventWarp(dst int32) {
 	if !ctrl.InParty() {
```
Hucaru: This needs to be event check

channel/script.go
Hucaru: Iterate over event playerIDs and lookup the player via ctrl.server

scripts/npc/2040044.js
```diff
@@ -5,15 +5,20 @@ var props = map.properties();
...
 } else if (plr.itemCount(key) >= 1) {
-    if (npc.sendYesNo("Incredible! You defeated Alishar and obtained the #t" + key + "#! Would you like to proceed to the bonus stage?")) {
-        plr.removeItemsByID(key, 1);
-        props.clear = true;
-        map.showEffect("quest/party/clear");
-        map.playSound("Party1/Clear");
-        map.portalEffect("gate");
-        plr.partyGiveExp(27200);
-        plr.partyWarp(922011000);
+    if (ctrl.isParticipantsOnMap(plr.mapID())) {
```
Hucaru: I can't see where this function is being exposed to the script wrapper
```
type scriptPlayerWrapper struct {
	plr    *Player
	server *Server
}
```

I think you need to expose this via a function e.g.:
```
func (ctrl *scriptPlayerWrapper) EventMembersOnMap(id int32) bool {
  if ctrl.player.event == nil {
    return false
  }

  return ctrl.plater.event.IsParticipantsOnMap(id)
}
func (ctrl *scriptPlayerWrapper) WarpEventMembers(id int32) {
  if ctrl.player.event != nil {
     ctrl.player.event.WarpPlayers(id)
  }
}

PR #241, opened 2026-01-16
Add standalone dev mode

ErwinsExpertise:
This pull request introduces a new "Dev Mode" (all-in-one) for simplified local development and testing, allowing all server types (login, world, channel, cashshop) to run in a single process. It also improves documentation, adds a unified dev configuration, and enhances shutdown and autosave robustness. Additionally, it makes the Prometheus metrics server lifecycle more robust and controllable.

The most important changes are:

**Dev Mode (All-in-One) Support:**
- Added a new `devServer` implementation (`server_dev.go`) that starts login, world, two channel, and cashshop servers in a single process, with coordinated startup, graceful shutdown, and signal handling. This enables simplified solo play and development workflows.
- Introduced support for the `dev` server type in `main.go`, including command-line flag help and dispatch logic. [[1]](diffhunk://#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L13-R13) [[2]](diffhunk://#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R35-R39)
- Added a unified `config_dev.toml` configuration file containing all necessary settings for all server types in dev mode.

**Documentation Updates:**
- Updated `.goreleaser.yaml` and `README.md` with clear instructions and explanations for Dev Mode, including quick start steps, features, and configuration guidance. [[1]](diffhunk://#diff-7326b55c062b0f46fe9e39aace0a25f4515cf206040fb91a6fd2cae839f5e826L93-R129) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R151-R194)

**Graceful Shutdown and Autosave Improvements:**
- Refactored `CheckpointAll` and `StartAutosave` in both `channel/server.go` and `cashshop/server.go` to accept a `context.Context`, supporting cancellation, timeouts, and improved logging for flush operations during shutdown. [[1]](diffhunk://#diff-767084af72220060b61c889aea587d211458143a5a229e2feb4736bacb8792cdL280-L289) [[2]](diffhunk://#diff-767084af72220060b61c889aea587d211458143a5a229e2feb4736bacb8792cdL307-R342) [[3]](diffhunk://#diff-420e5bee5b2808d68cdc3610d0be9f673ad247cc9284216b8c8bf2bc8321f883L106-L115) [[4]](diffhunk://#diff-420e5bee5b2808d68cdc3610d0be9f673ad247cc9284216b8c8bf2bc8321f883L133-R168)
- Updated server shutdown logic in `server_channel.go` and `server_cashshop.go` to pass the appropriate context to `CheckpointAll`. [[1]](diffhunk://#diff-2bebcd593322d19353442efbf0c8705f78b472ff446f35dadc32ed3a7b139951L103-R103) [[2]](diffhunk://#diff-62e2145f1fd4e09c0b57a0f5aca07d952bcf45e058b3536176fa22dce4c6e35cL94-R94)

**Metrics Server Lifecycle Enhancements:**
- Improved Prometheus metrics server handling in `common/metrics.go` by ensuring it starts only once (`sync.Once`), can be gracefully stopped, and avoids port conflicts in multi-server scenarios. [[1]](diffhunk://#diff-da6ba0ffc8743dea9032a01f7d5047b164f30731b2e6a99dfc691f3bb83e06ecR4-R8) [[2]](diffhunk://#diff-da6ba0ffc8743dea9032a01f7d5047b164f30731b2e6a99dfc691f3bb83e06ecR23-R63)

These changes collectively make local development, testing, and server management more robust and user-friendly, especially for new contributors or those running the project in solo/development environments.

### DIFF
--- a/server_dev.go
+++ b/server_dev.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"strconv"
 	"sync"
 	"syscall"
 	"time"
@@ -86,37 +87,37 @@ func (ds *devServer) run() {
 	// Give world server time to connect to login
 	time.Sleep(2 * time.Second)
 
-	// Start channel 1 server
-	ds.wg.Add(1)
-	go func() {
-		defer ds.wg.Done()
-		defer func() {
-			if r := recover(); r != nil {
-				log.Println("Channel server panic:", r)
-			}
-		}()
-		cs1 := newChannelServer(ds.configFile)
-		ds.channelServers = append(ds.channelServers, cs1)
-		cs1.run()
-	}()
+	ds.channelServers = make([]*channelServer, *channelPtr)
 
-	// Start channel 2 server
-	ds.wg.Add(1)
-	go func() {
-		defer ds.wg.Done()
-		defer func() {
-			if r := recover(); r != nil {
-				log.Println("Channel server panic:", r)
-			}
-		}()
-		cs2 := newChannelServer(ds.configFile)
-		cs2.config.ListenPort = "8686"
-		ds.channelServers = append(ds.channelServers, cs2)
-		cs2.run()
-	}()
+	for i := 0; i < *channelPtr; i++ {
+		ch := i
 
-	// Give channel server time to connect to world
-	time.Sleep(2 * time.Second)
+		cs := newChannelServer(ds.configFile)
+		cs.config.ListenPort = strconv.Itoa(8685 + ch)
+		ds.channelServers[ch] = cs
+
+		ds.wg.Add(1)
+		go func(s *channelServer) {
+			defer ds.wg.Done()
+			defer func() {
+				if r := recover(); r != nil {
+					log.Println("Channel server panic:", r)
+				}
+			}()
+			s.run()
+		}(cs)
+
+		select {
+		case <-cs.Ready():
+			log.Printf("Channel %d is ready (port %s)\n", ch, cs.config.ListenPort)
+		case <-time.After(30 * time.Second):
+			log.Printf("Timed out waiting for channel %d to become ready\n", ch)
+			ds.shutdown()
+			return
+		case <-ds.ctx.Done():
+			return
+		}
+	}
 
 	// Start cashshop server
 	ds.wg.Add(1)
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -99,7 +99,7 @@ release:
     3. Set up your MySQL database (see README.md)
     4. Run in dev mode:
     ```bash
-    ./Valhalla -type dev -config config_dev.toml
+    ./Valhalla -type dev -config config_dev.toml -channels 4
     ```
     
     Dev mode features:
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -35,11 +35,12 @@ VALHALLA_LOGIN_CLIENTLISTENADDRESS=0.0.0.0
 
 All server executables accept the following flags:
 
-| Flag | Required | Description | Example |
-|------|----------|-------------|---------|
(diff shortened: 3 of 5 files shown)


Hucaru:
Useful for dev. It should start 2 channels though to test things like party syncs etc

cashshop/server.go
```diff
@@ -103,16 +103,34 @@ func (server *Server) ClientDisconnected(conn mnet.Client) {
...
 
 // CheckpointAll now uses the saver to flush debounced/coalesced deltas for every player.
-func (server *Server) CheckpointAll() {
+func (server *Server) CheckpointAll(ctx context.Context) {
 	if server.dispatch == nil {
 		return
 	}
+
 	done := make(chan struct{})
-	server.dispatch <- func() {
```
Hucaru, line 111: I take it this was blocking?

ErwinsExpertise:
> Useful for dev. It should start 2 channels though to test things like party syncs etc

Good idea! Added a 2nd channel.

cashshop/server.go
ErwinsExpertise: Yes it was. Ran a pprof to figure out why it couldn't shut down and this was blocking.

server_dev.go
```diff
@@ -95,8 +95,24 @@ func (ds *devServer) run() {
...
 		}()
-		ds.channelServer = newChannelServer(ds.configFile)
-		ds.channelServer.run()
+		cs1 := newChannelServer(ds.configFile)
+		ds.channelServers = append(ds.channelServers, cs1)
+		cs1.run()
+	}()
+
+	// Start channel 2 server
+	ds.wg.Add(1)
```
Hucaru: Something like the following should be more flexible
```
ds.channelServers = make([]channelServer, config.ChannelCount)

for i := 0; i < config.ChanneCount; i++ {
	ds.wg.Add(1)
	go func() {
		defer ds.wg.Done()
		defer func() {
			if r := recover(); r != nil {
				log.Println("Channel server panic:", r)
			}
		}()
		cs := newChannelServer(ds.configFile)
		cs.config.ListenPort = strconv.Itoa(config.ChannelStartPort + i)
		ds.channelServers[i] = cs // Need to index directly to avoid 2 threads apending at same time
		cs.run()
	}()
}
```

ErwinsExpertise: Good idea. I added it as a command flag instead since this is really only meant for running locally,.

server_dev.go
```diff
@@ -86,37 +87,37 @@ func (ds *devServer) run() {
...
+			defer ds.wg.Done()
+			defer func() {
+				if r := recover(); r != nil {
+					log.Println("Channel server panic:", r)
+				}
+			}()
+			s.run()
+		}(cs)
+
+		select {
```
Hucaru, line 110: I don't think you need to check if it's ready. The servers should still connect to one another regardless of startup order, I'm not a fan of adding a channel in the main server code for dev server handling.

ErwinsExpertise: @Hucaru the reason for ready check is because they all try to mutate(unmarshal causes it) the files such as drops json at the same time which crashes the channel.  With the ready check in place it will not move forward in the loop until the channel is ready preventing those errors. We could add a sleep, but that’s pretty hacky and load time could change based on tons of variables. 

I actually quite like the readiness check honestly. It is adding another channel sure, but it gets closed immediately when the channel starts listening so it doesn’t hang around. 

There’s definitely a few different deployment strategies that could make use of a proper ready check, but this is just the immediate use-case.

Hucaru: Ok, for now lets do it this way and change it as part of a wider code cleanup.

Hucaru (approved):

PR #250, opened 2026-01-20
add anticheat system

ErwinsExpertise:
This pull request introduces a comprehensive anti-cheat system to the server, adding detection and enforcement for various forms of cheating, such as movement hacks, attack speed hacks, invalid trades, and more. It implements a new `AntiCheat` module with violation tracking, auto-banning, ban escalation, and ban history, and integrates these checks into player actions throughout the codebase. Additionally, it adds GM commands for manual ban management and provides mechanisms for kicking banned players.

**Key changes:**

**Anti-cheat system implementation:**
- Added a new `AntiCheat` module (`anticheat/anticheat.go`) that tracks violations, failed authentications, and issues bans (temporary or permanent) based on configurable thresholds. It supports escalation for repeated temporary bans, maintains ban history, and provides helper methods for various cheat detections.
- Integrated the anti-cheat system into the server's lifecycle: initialized in `Server.Initialise`, started periodic cleanup, and set up a callback to kick banned accounts. [[1]](diffhunk://#diff-767084af72220060b61c889aea587d211458143a5a229e2feb4736bacb8792cdR13) [[2]](diffhunk://#diff-767084af72220060b61c889aea587d211458143a5a229e2feb4736bacb8792cdR50) [[3]](diffhunk://#diff-767084af72220060b61c889aea587d211458143a5a229e2feb4736bacb8792cdR164-R171)

**Ban enforcement and player management:**
- Added a `KickAccount` method to the server to disconnect banned players, and a `Kick` method to the `Player` struct for clean disconnection. These are used both by the anti-cheat system and GM commands. [[1]](diffhunk://#diff-767084af72220060b61c889aea587d211458143a5a229e2feb4736bacb8792cdR208-R240) [[2]](diffhunk://#diff-7c17860a8d1d10607463092eb6652bafd66b1f34027f14eeb42fa052a1385971R1753-R1761)

**Cheat detection integration:**
- Added anti-cheat checks to player actions:
  - Movement: Detects teleport hacks and suspicious movement distances. [[1]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR409-R425) [[2]](diffhunk://#diff-b46fadb5eadf53ca9fa0cbed7a6639cc44cb36716c2f60d5a8248dc10b6a7001L188-R214)
  - Item usage: Tracks invalid item usage.
  - Skills: Detects skill abuse, excessive attack speed, and excessive damage for melee, ranged, and magic skills. [[1]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL2466-R2499) [[2]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL2485-R2512) [[3]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR2529-R2532) [[4]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL2536-R2580) [[5]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR2620-R2622) [[6]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL2596-R2637)
  - Trades: Detects invalid trades, such as overflow or selling items not owned. [[1]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR3127-R3129) [[2]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR3152-R3154) [[3]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR3177-R3179) [[4]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fR3191-R3193)

**Game master (GM) commands:**
- Added new GM commands for ban management: `/ban`, `/unban`, and `/banhistory`, allowing GMs to manually ban/unban players and view ban history directly in-game.

**Miscellaneous:**
- Updated imports and minor code adjustments to support the new anti-cheat features.

These changes significantly enhance the server's ability to detect, prevent, and respond to cheating, providing both automated and manual tools for maintaining fair gameplay.

### DIFF
--- a/login/handlers.go
+++ b/login/handlers.go
@@ -493,13 +493,15 @@ func (server *Server) handleDeleteCharacter(conn mnet.Client, reader mpacket.Rea
 
 	if charCount != 1 {
 		if server.ac != nil {
-			err = server.ac.IssueBan(0, 24, "Excessive failed login attempts", conn.String(), conn.GetHWID())
+			err = server.ac.IssueBan(0, 24, "Attempted to delete character not associated with account", conn.String(), conn.GetHWID())
 			if err != nil {
 				log.Println(err)
 			}
 		}
-		hacking = true
-		conn.Send(packetLoginDeleteCharacter(charID, deleted, hacking))
+		err := conn.Close()
+		if err != nil {
+			log.Println(err)
+		}
 		return
 	}
 
@@ -532,7 +534,7 @@ func (server *Server) handleSelectCharacter(conn mnet.Client, reader mpacket.Rea
 	if err != nil {
 		log.Println(err)
 		if server.ac != nil {
-			err = server.ac.IssueBan(0, 24, "Excessive failed login attempts", conn.String(), conn.GetHWID())
+			err = server.ac.IssueBan(0, 24, "Attempted to select character not associated with account", conn.String(), conn.GetHWID())
 			if err != nil {
 				log.Println(err)
 			}


anticheat/anticheat.go
```diff
@@ -0,0 +1,334 @@
...
+	ac.post(func() {
+		for _, id := range identifiers {
+			delete(ac.failedAuth, id)
+		}
+	})
+}
+
+// IssueBan creates a temporary ban (hours=0 means permanent)
+func (ac *AntiCheat) IssueBan(accountID int32, hours int, reason string, ip, hwid string) error {
+	var banEnd interface{}
```
Hucaru: Multiple: Why is this an interface type instead of a time related type?

channel/commands.go
```diff
@@ -1720,6 +1720,72 @@ func (server *Server) gmCommand(conn mnet.Client, msg string) {
...
+		if len(command) >= 4 {
+			reason = strings.Join(command[3:], " ")
+		}
+
+		if server.ac != nil {
+			server.ac.IssueBan(target.Conn.GetAccountID(), hours, reason, "", "")
+			conn.Send(packetMessageRedText(fmt.Sprintf("Banned %s for %d hours", targetName, hours)))
+		}
+
+	case "unban":
```
Hucaru, line 1753: The next two commands should work off character name and account ID. Otherwise it requires a GM/Admin to have database read access.

channel/damage_calc.go
```diff
@@ -28,36 +27,67 @@ type CalcHitResult struct {
 	IsValid      bool
 }
 
-type Roller struct {
-	rollIndex int
-	rolls     []uint32
+type DamageRngBuffer struct {
+	raw [constant.DamageRngBufferSize]uint32
```
Hucaru: If the struct only has one field then it might be worth changing to a typedef

channel/damage_calc.go
```diff
@@ -28,36 +27,67 @@ type CalcHitResult struct {
 	IsValid      bool
 }
 
-type Roller struct {
-	rollIndex int
-	rolls     []uint32
+type DamageRngBuffer struct {
```
Hucaru: Is the purpose of this to simulating the client "randomness"? A comment describing the technicalities of this would be useful

channel/handlers_client.go
```diff
@@ -406,6 +406,23 @@ func (server Server) playerMovement(conn mnet.Client, reader mpacket.Reader) {
...
+		lastFrag := moveData.frags[len(moveData.frags)-1]
+
+		dx := float64(lastFrag.x - firstFrag.x)
+		dy := float64(lastFrag.y - firstFrag.y)
+		distance := math.Sqrt(dx*dx + dy*dy)
+
+		if distance > 1000 {
+			reason := fmt.Sprintf("Suspicious movement: %.0f pixels", distance)
+			log.Println("Teleport hack detected:", reason, "accountID:", plr.accountID)
+			if server.ac != nil {
```
Hucaru: You have already checked if this is not nil

channel/handlers_client.go
```diff
@@ -406,6 +406,23 @@ func (server Server) playerMovement(conn mnet.Client, reader mpacket.Reader) {
...
+				server.ac.CheckMovement(plr.accountID, int16(distance))
```
Hucaru: Check movement also checks the distance value is > 1000. This should ideally only be performed in one location

channel/handlers_client.go
```diff
@@ -1364,6 +1381,9 @@ func (server Server) playerUseInventoryItem(conn mnet.Client, reader mpacket.Rea
 	item, err := plr.takeItem(itemid, slot, 1, 2)
 	if err != nil {
 		log.Println(err)
+		if server.ac != nil {
+			server.ac.CheckInvalidItem(plr.accountID)
```
Hucaru: These names can be a bit missleading as the intuitive reaction to them is that they are performin the detection. 

They should probably convey that you are checking the re-offence threshold in the function name.

channel/damage_calc.go
ErwinsExpertise: I recently learned how GMS does it:
https://gist.github.com/ThatMapleDev/9e137530009be8460a03f5b3f8283c2e#4-the-7-random-heartbeat-model

Figured it would be a good idea to match it.

channel/handlers_client.go
```diff
@@ -2499,6 +2526,10 @@ func validateAndApplyCriticals(plr *Player, data *attackData, results [][]CalcHi
 
 				log.Printf("Capped excessive damage from player %s (ID: %d): client=%d -> capped=%d, max=%.0f, skill=%d",
 					plr.Name, plr.ID, result.ClientDamage, maxAllowed, result.MaxDamage, data.skillID)
+
+				if server.ac != nil {
+					server.ac.CheckDamage(plr.accountID, result.ClientDamage, int32(result.MaxDamage))
```
Hucaru: Is going over damage cap something that can be done with perfect gear and hardest hitting skill?

channel/player.go
```diff
@@ -1750,6 +1750,15 @@ func (d Player) Logout() {
 
 }
 
+func (d *Player) Kick() {
+	if d == nil || d.Conn == nil {
+		return
+	}
+
+	d.Send(packetChangeChannel([]byte{127, 0, 0, 1}, 0))
+	_ = d.Conn.Close()
```
Hucaru, line 1758: Is this needed? The connection should close once the client disconnects

channel/server.go
```diff
@@ -195,6 +205,39 @@ func (server *Server) SendLostWorldConnectionMessage() {
...
+	do := func() {
+		server.players.observe(func(plr *Player) {
+			if plr == nil || plr.Conn == nil {
+				return
+			}
+			if plr.Conn.GetAccountID() != accountID {
+				return
+			}
+			if message != "" {
+				plr.Send(packetMessageNotice(message))
```
Hucaru: World wide message?

login/handlers.go
```diff
@@ -47,6 +49,24 @@ func (server *Server) HandleClientPacket(conn mnet.Client, reader mpacket.Reader
 func (server *Server) handleLoginRequest(conn mnet.Client, reader mpacket.Reader) {
 	username := reader.ReadString(reader.ReadInt16())
 	password := reader.ReadString(reader.ReadInt16())
+
+	// Try to read HWID (6 bytes after password) - simplified
+	hwid := ""
+	if reader.ReadInt16() >= 6 { // Check remaining bytes
+		hwidBytes := make([]byte, 6)
+		for i := 0; i < 6; i++ {
+			hwidBytes[i] = reader.ReadByte()
```
Hucaru: There's a ReadBytes() method

login/handlers.go
```diff
@@ -47,6 +49,24 @@ func (server *Server) HandleClientPacket(conn mnet.Client, reader mpacket.Reader
 func (server *Server) handleLoginRequest(conn mnet.Client, reader mpacket.Reader) {
 	username := reader.ReadString(reader.ReadInt16())
 	password := reader.ReadString(reader.ReadInt16())
+
+	// Try to read HWID (6 bytes after password) - simplified
+	hwid := ""
+	if reader.ReadInt16() >= 6 { // Check remaining bytes
```
Hucaru: Can this be greater than 6?

login/handlers.go
```diff
@@ -110,6 +160,24 @@ func (server *Server) handleLoginRequest(conn mnet.Client, reader mpacket.Reader
...
+			common.DB.Exec("UPDATE accounts SET hwid = ? WHERE accountID = ?", hwid, accountID)
+		}
+		if server.ac != nil {
+			ip := ""
+			if host, _, err := net.SplitHostPort(conn.String()); err == nil {
+				ip = host
+			} else {
+				ip = conn.String()
+			}
+			server.ac.ClearAuth(
```
Hucaru, line 178: I have never seen a system that bans for failed login attempts. At most the account should get locked and require an unlock via a website interaction.

mnet/conn.go
```diff
@@ -144,6 +148,18 @@ func (bc *baseConn) String() string {
...
-	close(bc.eSend)
+	bc.closeOnce.Do(func() {
+		bc.closed = true
+		if bc.eSend != nil {
+			close(bc.eSend)
+		}
+	})
+}
+
+func (bc *baseConn) Close() error {
```
Hucaru, line 159: Tied to my earlier comment on kicking player

login/handlers.go
ErwinsExpertise: Hmm, we can add in an account lock rather than ban, but there's also no website support so I think I'd need to add in an unlock gm command too...?

channel/handlers_client.go
ErwinsExpertise: Nope, we calculate the max change for the player based on the known damage formulas, and we even provide a 15% variance. If the provided damage is exceeding those limits then it is most likely hacking.

channel/player.go
ErwinsExpertise: I would prefer closing the connection forcefully rather than issuing a change channel packet. The packet was just to speed up the process, but that packet could always get intercepted from a network level and prevent it from coming through to the client. Closing the connection forcefully is the only guaranteed method to kick them.

mnet/conn.go
ErwinsExpertise: https://github.com/Hucaru/Valhalla/pull/250#discussion_r2714172234

login/handlers.go
ErwinsExpertise: This was a little odd so refactored to use skip, and readbytes instead. The sizes never change, they are fixed.

login/handlers.go
```diff
@@ -50,15 +50,10 @@ func (server *Server) handleLoginRequest(conn mnet.Client, reader mpacket.Reader
...
-	hwid := ""
-	if reader.ReadInt16() >= 6 { // Check remaining bytes
-		hwidBytes := make([]byte, 6)
-		for i := 0; i < 6; i++ {
-			hwidBytes[i] = reader.ReadByte()
-		}
-		hwid = fmt.Sprintf("%02X%02X%02X%02X%02X%02X", hwidBytes[0], hwidBytes[1], hwidBytes[2], hwidBytes[3], hwidBytes[4], hwidBytes[5])
-	}
+	// Try to read HWID
+	reader.Skip(6)
```
Hucaru, line 56: Are these not part of HWID? I assumed the ID was a mac address.

login/handlers.go
Hucaru: It would be better if instead of banning the user it set a field in the database for the account to lock it. This would then be up to the privaste server on how they want to manage unlocking accounts (probably via some website password reset utility)

channel/player.go
Hucaru: In that case I don't think you need to send the change channel. Terminating the socket will be immediate for the server. The hacker's client might not close imediately but from the server's perspective it doesn't matter.

login/handlers.go
ErwinsExpertise: It is the machine ID, the first 6 bytes just seem to be 0's so I assume it's padding.

login/handlers.go
ErwinsExpertise: Alright added in `isLocked`. We still ban IP/HWID's globally(for 1 hour) for failed attempts, but we also lock the account instead of banning it.

Hucaru (dismissed):

login/handlers.go
```diff
@@ -91,7 +118,29 @@ func (server *Server) handleLoginRequest(conn mnet.Client, reader mpacket.Reader
...
+
+			exceeded := server.ac.TrackFailedAuth(ipKey) || server.ac.TrackFailedAuth(hwidKey) || server.ac.TrackFailedAuth(userKey)
+			if exceeded {
+				// Lock Account
+				err := server.ac.LockAccount(accountID)
+				if err != nil {
+					log.Println(err)
+				}
+				// Ban IP & HWID globally
+				err = server.ac.IssueBan(0, 24, "Excessive failed login attempts", ip, hwid)
```
Hucaru: This could lead to the situation where someone has forgotten their password and is now unable to log into other accounts they might have

ErwinsExpertise: Okay modified it to be this:
```
if exceeded && strings.Compare(hwid, lastHwid) != 0 {
```

We set the `hwid` for the account in DB on successful login, and `hwid` is generally a better tracker as IP's can change all the time for various reason, but the machine ID will rarely change. If they exceed attempts *and* the machine ID doesn't match the last known ID then we can generally assume it has a higher likelihood of being a bad actor, and in that case we do want to ban them to mitigate brute-force attempts.

login/handlers.go
```diff
@@ -127,7 +128,7 @@ func (server *Server) handleLoginRequest(conn mnet.Client, reader mpacket.Reader
 			hwidKey := fmt.Sprintf("hwid:%s", hwid)
 
 			exceeded := server.ac.TrackFailedAuth(ipKey) || server.ac.TrackFailedAuth(hwidKey) || server.ac.TrackFailedAuth(userKey)
-			if exceeded {
+			if exceeded && strings.Compare(hwid, lastHwid) != 0 {
```
Hucaru: What you had here was fine. I was thinking more of removing the ip and hardware id ban. Those should only be reserved for someone who is hacking in the game or packet manipulation. The only punishment for not correctly guessing a password should be to lock the account and prevent any game logins until the account is unlocked.

login/handlers.go
Hucaru, line 544: If char count is 0 then that means they provided an ID they do not own which should only be possible via packet manipulation and they should be IP and hardware banned.

Hucaru, line 1: There are a lot of panics when handling errors and I think we should change them to just return from function as we don't want to crash the login server as other usrs might be logged in and account could potentially be stuck in a logged in state when the server restarts.

login/handlers.go
ErwinsExpertise: The idea is to mitigate brute-force attempts, and it’s only a 1 hour. Do you believe that is too harsh for what would otherwise appear to be malicious? If we just lock the account then the potential bad actor can just move on to a different account. Assuming it’s someone who forgot their password and is logging on from the same PC as usual it will not negatively impact them.

We can remove the ban, but I think this would help with mitigation efforts for server owners.

login/handlers.go
ErwinsExpertise: @Hucaru i don’t see any panics, which do you mean?

ErwinsExpertise: Ah, the existing ones. I went ahead and made those changes.

login/handlers.go
```diff
@@ -485,8 +492,15 @@ func (server *Server) handleDeleteCharacter(conn mnet.Client, reader mpacket.Rea
...
 	if charCount != 1 {
-		log.Println(conn.GetAccountID(), "attempted to delete a character they do not own:", charID)
+		if server.ac != nil {
+			err = server.ac.IssueBan(0, 24, "Excessive failed login attempts", conn.String(), conn.GetHWID())
+			if err != nil {
+				log.Println(err)
+			}
+		}
 		hacking = true
+		conn.Send(packetLoginDeleteCharacter(charID, deleted, hacking))
```
Hucaru: Do we need to remove the character from the client if they have been banned?

login/handlers.go
```diff
@@ -485,8 +492,15 @@ func (server *Server) handleDeleteCharacter(conn mnet.Client, reader mpacket.Rea
 	deleted := false
 
 	if charCount != 1 {
-		log.Println(conn.GetAccountID(), "attempted to delete a character they do not own:", charID)
+		if server.ac != nil {
+			err = server.ac.IssueBan(0, 24, "Excessive failed login attempts", conn.String(), conn.GetHWID())
```
Hucaru: Ban reason incorrect?

login/handlers.go
ErwinsExpertise: True, how do you feel about just closing the connection @Hucaru ?

login/handlers.go
ErwinsExpertise: Copy pasta, oops

PR #255, opened 2026-01-23
Added mesos tax

### LINKED ISSUE #248: Implement trade tax
Trade is taxed at certain % of mesos. Needs to be added to:
https://github.com/Hucaru/Valhalla/blob/master/channel/room_trade.go#L202-L208

Hucaru:
Implements mesos trade tax and cleaned up mesos and item exchange

Closes #248

### DIFF
--- a/channel/room_trade.go
+++ b/channel/room_trade.go
@@ -155,22 +155,15 @@ func (r *tradeRoom) completeTrade() {
 	}
 
 	var undo []func()
-	defer func() {
-		if r.finalized {
-			return
-		}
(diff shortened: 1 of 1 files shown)


channel/room_trade.go
```diff
@@ -155,58 +155,53 @@ func (r *tradeRoom) completeTrade() {
 	}
 
 	var undo []func()
-	defer func() {
-		if r.finalized {
-			return
-		}
-		for i := len(undo) - 1; i >= 0; i-- {
```
Hucaru, line 162: This can only get triggered at item exchange, so just inlined there

PR #257, opened 2026-01-26
Implement Pickpocket meso drops and fix Mesos Explosion 

ErwinsExpertise:
This pull request implements a complete and accurate server-side calculation for the Meso Explosion skill, ensuring that damage is based on the actual mesos consumed and using the official formula. It also adds a correct implementation for the Pickpocket skill, allowing mesos to drop from mobs when the skill procs. Additionally, constants for Meso Explosion's damage formula are now centralized, and client-reported damages are validated to prevent abuse.

**Meso Explosion Skill Calculation:**

* Server-side calculation for Meso Explosion now uses the actual mesos consumed, applies the official formula, and validates client-reported damages within a tolerance to prevent exploits (`channel/damage_calc.go`, `channel/handlers_client.go`, `constant/damage.go`) [[1]](diffhunk://#diff-010c97e3c360e16e2623a888109e921393a5ac389941c75b8281549b289be42bL203-R203) [[2]](diffhunk://#diff-010c97e3c360e16e2623a888109e921393a5ac389941c75b8281549b289be42bL296-R362) [[3]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL2643-R2664) [[4]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL2656-R2707) [[5]](diffhunk://#diff-4d034f01a8cb51793c70f45f437dc141ce438df213d14b8e2db724aa03ac1be6R63-R71).
* Meso Explosion formula constants are defined in `constant/damage.go` for maintainability and clarity.

**Pickpocket Skill Implementation:**

* Added logic to proc Pickpocket on mob damage, rolling for meso drops according to skill level and dropping mesos at the mob's position for the player (`channel/pools.go`).

**Client Damage Validation:**

* Client-reported damage for Meso Explosion is now checked against a calculated maximum with a variance tolerance, replacing it with the server-calculated value if outside the allowed range (`channel/damage_calc.go`, `channel/handlers_client.go`) [[1]](diffhunk://#diff-010c97e3c360e16e2623a888109e921393a5ac389941c75b8281549b289be42bL296-R362) [[2]](diffhunk://#diff-58f47af606b2ed1ca1742a51a8b1f41aa34bd82ecca1e605b23c52cb9fe4cd8fL2656-R2707).

### DIFF
--- a/channel/pools.go
+++ b/channel/pools.go
@@ -351,8 +351,7 @@ func (pool *lifePool) mobDamaged(poolID int32, damager *Player, dmg ...int32) {
 								mesoAmount = 5000
 							}
 
-							dropPos := pool.instance.calculateFinalDropPos(v.pos)
-							pool.dropPool.createDrop(dropSpawnNormal, dropTimeoutNonOwner, mesoAmount, dropPos, true, damager.ID, 0)
+							pool.dropPool.createDrop(dropSpawnNormal, dropTimeoutNonOwner, mesoAmount, v.pos, true, damager.ID, 0)
 						}
 					}
 				}


channel/damage_calc.go
```diff
@@ -293,11 +293,73 @@ func (calc *DamageCalculator) CalculateHit(
...
+		// If no mesos found, damage should be 0
+		if totalMesos == 0 {
+			result.MinDamage = 0
+			result.MaxDamage = 0
+			result.ExpectedDmg = 0
+			result.IsValid = (result.ClientDamage == 0)
+			return true
+		}
+		
+		// Calculate damage using the correct formula
```
Hucaru, line 349: This calculation step is happening multiple times. First time here and then a second time in the mesos explosion attack handler. This validation should probably adjust the damage if it's gone above max damage and the mesos explosion handler should not need to recheck if damage conforms to formula.

Hucaru: Thinking about this further I think it might be better to just do the validation here. If the validation passes then the damage can be corrected in the attack handler.

ErwinsExpertise: Removed duplicate calc

channel/pools.go
```diff
@@ -327,6 +327,35 @@ func (pool *lifePool) mobDamaged(poolID int32, damager *Player, dmg ...int32) {
...
+							if mesoAmount < 1 {
+								mesoAmount = 1
+							}
+
+							if mesoAmount > 5000 {
+								mesoAmount = 5000
+							}
+
+							dropPos := pool.instance.calculateFinalDropPos(v.pos)
+							pool.dropPool.createDrop(dropSpawnNormal, dropTimeoutNonOwner, mesoAmount, dropPos, true, damager.ID, 0)
```
ErwinsExpertise: @Hucaru for some reason this drop is raining from the sky vs dropping from monster pos as normal drops do. Any idea what's wrong?

channel/damage_calc.go
```diff
@@ -293,11 +293,73 @@ func (calc *DamageCalculator) CalculateHit(
 	return result
 }
 
-func (calc *DamageCalculator) handleSpecialSkillDamage(result *CalcHitResult, mob *monster, rngBuf *DamageRngBuffer, hitIdx int) bool {
+func (calc *DamageCalculator) handleSpecialSkillDamage(result *CalcHitResult, mob *monster, info *attackInfo, rngBuf *DamageRngBuffer, hitIdx int) bool {
 	str := float64(calc.player.str)
 	dex := float64(calc.player.dex)
 	luk := float64(calc.player.luk)
 
+	if skill.Skill(calc.skillID) == skill.MesoExplosion {
```
Hucaru: I think this if statement along with the next one(s) would be better represented as a switch statement now

channel/pools.go
Hucaru: I'm not sure I follow exactly. Is it acting like Ludi bonus stage drop? Or is the drop starting in the sky instead of the ground? Are you able to post a screenshot so I can see the problem?

channel/pools.go
```diff
@@ -327,6 +327,35 @@ func (pool *lifePool) mobDamaged(poolID int32, damager *Player, dmg ...int32) {
...
+							mesoAmount := int32(float64(dmgSum) * mesoPercentage)
+							if mesoAmount < 1 {
+								mesoAmount = 1
+							}
+
+							if mesoAmount > 5000 {
+								mesoAmount = 5000
+							}
+
+							dropPos := pool.instance.calculateFinalDropPos(v.pos)
```
Hucaru: This might be the issue. It should be `v.pos`. createDrop will do the final drop calculation for you on this line https://github.com/Hucaru/Valhalla/blob/master/channel/pools.go#L1048

channel/pools.go
ErwinsExpertise: Yeah, it falls from the top of the screen. 
https://prnt.sc/6GrltJCe5E7V

channel/pools.go
ErwinsExpertise: I added this after hoping it was the foothold, but it does not seem to be the issue.

channel/pools.go
Hucaru: can you test with this set to zero https://github.com/Hucaru/Valhalla/blob/master/channel/field.go#L1041 ? I suspect this is passing a cross product check in the foothold above code for a foothold it shouldn't

ErwinsExpertise: I attempted with that and it's still raining. The drop seems to be making it to the correct foothold, however, the animation isn't right.

ErwinsExpertise: https://prnt.sc/qs0yL-inQAzF

It looks like the drop is coming from a single mob and spreading to all that are hit? I'm assuming this mob is likely the first in the slice, but we are using the indexed position so that doesn't really make sense?

Hucaru: I think this might happen when there is a mob that shares the same spawn id as the character on the map.

A quick check would be to to have 2 characters with low id's e.g. 1 and 2 and see if when pickpocket is enabled the drops originate from different monsters. When creating the mesos from pickpocket print the mob spawn id you are hitting and find the origin mob and confirm it has the same id as the character. 

A quick hack solution would be to start indexing character id's in the database at 1000 cycle mob IDs up to 1000 instead of max int32. This is very dirty and should just be a final resort.

I think the reason this doesn't happen when a mob dies is because we remove the mob before me make the drop.

Hucaru: If this ends up being the issue don't fix in this PR as player character storage in server might need a rework to be part of life pool

ErwinsExpertise: Yeah it seems to mostly work fine with any character past the first index.

Hucaru: Ok, then leave it for now and raise a bug ticket to keep track of it. Exact conditions need to be determined before a proper fix can be determined.

channel/pools.go
Hucaru: This line needs to be removed it's redundant

